PR #172, opened 2015-11-09
3.2.1 devel travis

### LINKED ISSUE #171: rake db:seed tries to create an existing user
The `rake db:seed` command attempts to create an admin user, but this user is already created in the migration stage. 

The solution (I think) is to comment out the user creation in the migration and leave it in the seed. I have a fix pending.

See:
https://github.com/rism-ch/muscat/blob/master/db/migrate/20140331105619_devise_create_users.rb#L5

https://github.com/rism-ch/muscat/blob/master/db/seeds.rb#L12

ahankinson:
This commit adds basic travis support for building Muscat.

It currently builds against ruby 1.9.3 and 2.1.1. 

This pull request also contains fixes for problems with the migrations (Refs #171). See the individual commits for changes.

No unit tests are run (yet).

Fixes #171


xhero:
Looks ok to me, @lpugin should I merge?

lpugin:
Looks good to me. Go ahead

lpugin:
Wait... You might want to see if everything work with the version in Gemfile.lock.

ahankinson:
Gemfile.lock was only changed because I added the sqlite3 gem, but then I realized that there are InnoDB-related things in the migrations, so I moved it back. So the Gemfile.lock is now effectively unchanged.

lpugin:
Gemfile is unchanged, but I think the Gemfile.lock specifies newer versions (I think). This is not an issue, but we need to make sure it works (in particular blacklight).

xhero:
I tested it and it seems ok, blacklight 5.16 is a very minor update

lpugin:
OK, thanks. Go ahead then.

PR #420, opened 2016-10-05
I18n spanish

drschoener:
Hope this is better now ...


xhero:
Very nice! thanks! There are just a couple "translation missing": is it OK if I add them into the file ALL CAPS so you can spot them?
If you want now you can also translate the marc editor label files, in config/editor_profiles/default/configurations/*Label*.yml
thanks!

PR #665, opened 2018-08-08
Merge 4.2 devel works into 4.3-devel

HirschSt:


xhero:
Seems very good to me! thanks!

PR #666, opened 2018-08-08
Merge 4.2 devel warnings into 4.3-devel

HirschSt:


xhero:
I have looked through the commits and it looks good, but I have a few thoughts. 
I think the code in https://github.com/rism-ch/muscat/pull/666/files#diff-8e4bed4cd54d462c8c686877120a9a9cR3 somewhat duplicates the existing MarcValidator class. I think we could merge the two so calling MarcValidator.validate will trigger all the existing validations plus the ones in this file.

Also I'm not completely sure I understand why the failed record is saved to a file (https://github.com/rism-ch/muscat/pull/666/files#diff-a95c389111276a83420de807554b61f1R22) and then reloaded. This looks to me like a duplication of the javascript validation. In my opinion we should implement all the checks also in javascript, so the user is prevented from saving invalid records, and then run the server-side one as a sanity check. The latter can be done with merging your code with the existing server side validation and notification (which I use for batch integrity checking, for example).

HirschSt:
Thanks for the review, Rodolfo!

IMO main problem for the existing JS validation is that you don't receive server (AR) side validation messages there. AR validation is only triggered in something :before_save callback.

My intention was not to touch the existing marcvalidator, so merging seems to be a natural thing to put all the related stuff together if they are fitting.

Basic idea for the filesave: form values are __not__ saved into the database in case of a AR validation error, but a redirect is triggered after the save action; so in case you want to restore all the form values (user did'nt want to fill out everything again) it's necessary to get the values somehow into the form again. 

BTW please postpone this pull request, I think we need a closer look (also from my side :wink: ) and rspecs.

HirschSt:
How about adding a generic validation controller? Then we could send an ajax request to this controller, with the json response as returning message. Also we could add a "validate" action to check the record without using the AR validation and we have all in JS; and we could drop the ugly save form to file. Any opinion?

PR #758, opened 2019-04-01
5.2 devel works

HirschSt:
this one adds some new fields to work, renaming LiturgicalFeasts to Events only on UI and some **new migration**

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190403102353) do
+ActiveRecord::Schema.define(version: 20190325082715) do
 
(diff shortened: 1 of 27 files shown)


db/schema.rb
```diff
@@ -312,6 +312,7 @@
     t.text "sub_topic"
     t.string "viaf"
     t.string "gnd"
+    t.string "task"
```
lpugin: Not too sure about the field name. Other ideas?

HirschSt: it's difficult to choose one which is not somehow meaningful with ruby (eg 'type' or 'function'), but I'm really open for any suggestions

HirschSt:
Make Undo of the renaming of LiturgicalFeast to Events as discussed and removed the migration to add "abstract" and "task" columns

Issue #713, opened 2018-12-11
Add method to find and check really all dependencies


HirschSt:
it looks like that the check_dependency method is looking for defined tables (see eg. https://github.com/rism-ch/muscat/blob/6483f82755f52fe3639bb3a00ee2cd575694d895/app/models/institution.rb#L248), it would be better to have a generic method without hardcoded checks, something like

`Institution.reflect_on_all_associations.map{|e| e.name}`

xhero:
+1 I like it! Will do asap

HirschSt:
some initial showcase with 5.2-devel-assoc, see https://github.com/rism-ch/muscat/commit/4cda68fdb9302d3b3ba0c1eb451a4f86718a4298

xhero:
While it is there, I think this is good for 6.2

PR #810, opened 2019-07-03
5.3 devel assoc

HirschSt:
This delicate, but important one is moving the check_dependencies from all the individual objects to a generic check, which is testing remaining external dependencies of an (marc)-object before destroying itself. 

The external references are collected directly from the marc configuration.

It's tested that all existent dependency_checks from 5.2 are preserved.

see https://github.com/rism-ch/muscat/issues/713 as reference


HirschSt:
BTW we can add this later for Muscat 6, but I thinks it's better to have some review now

PR #821, opened 2019-08-15
adding movement number and title

kepper:


kepper:
for the MARC codes, please check if this is really functional, or what comes out of this…

xhero:
Danke schön! I will run a test for the codes and see

PR #842, opened 2019-08-28
Changed composite configuration

HirschSt:


app/views/editor/_group_by_subfield.html.erb
```diff
@@ -22,11 +22,12 @@
 <%- end %>
 </dl>
 
+<% unless @item.record_type == 11 %>
```
lpugin: Can we think about another way to do this? Maybe with something more explicit in the configuration?

HirschSt: Yes, I am not glad with this too. But otherwise the complex structure of the configuration will be blowed up again by another element. Anyway: Proposals are welcome..

HirschSt: I have made an issue with that and removed the lines with the recent commit, so we have time to find a better solution

PR #854, opened 2019-10-04
Add sqlite3 gem

fjorba:
sqlite3 is used for testing db.  Add the requirement. 

The error raises, at least, when dropping and recreating db after a Muscat upgrade (not strictly needed, but in my case I wanted to clear it anyway).

```
~/download/muscat/muscat-5.2.1$ rails db:drop

Dropped database 'muscat_development'
Dropped database 'db/test.sqlite3'

~/download/muscat/muscat-5.2.1$ rails db:create
Created database 'muscat_development'
Specified 'sqlite3' for database adapter, but the gem is not loaded. Add `gem 'sqlite3'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).
Couldn't create database for {"adapter"=>"sqlite3", "database"=>"db/test.sqlite3", "pool"=>5, "timeout"=>5000}
rails aborted!
Gem::LoadError: Specified 'sqlite3' for database adapter, but the gem is not loaded. Add `gem 'sqlite3'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,9 @@ gem 'rails', '>= 5.1', '< 5.2'
 # https://github.com/rails/rails/issues/21544
 gem 'mysql2'
 
+# Use sqlite3 for testing db
+gem 'sqlite3'
+
 # Use SCSS for stylesheets
 gem 'sass-rails'#, '~> 4.0.0'
 


HirschSt:
Please check if this really is needed; for testing purposes we are using rspec and factories.

Probably your error is a result from calling the default (sample_)database.yml:https://github.com/rism-ch/muscat/blob/d13850fe311f73cfb502416238756f54a45c1d5a/config/sample_database.yml#L29

and might be solved by calling the db:create in a rails environment, eg 'production'

BTW: sqlite3 maybe not 100% compatible with the current mysql adapter at all

fjorba:
Probably you are right.  The file database.yml has the three entries (development, test and production), and I thought that, in Rails, exporting RAILS_ENV=development (in my case), would be enough so that Rails only picks up the approppiate entry, and it doesn't.  After commenting out those entries the error has gone.

Closing the request, thanks again.

> BTW: sqlite3 maybe not 100% compatible with the current mysql adapter at all

I'm very much interested in making Muscat sqlite3 compatible, but I'll leave that for later.

PR #928, opened 2020-04-29
Read default editor profiles yml

fjorba:
If I understand well the Editor Configuration documentation, if no specific config/editor_profiles/#{profile_name}/profiles.yml exist, it should use the one in config/editor_profiles/default/.  Take it into account.

Otherwise, the application breaks and gives a traceback.


xhero:
Yes I think we never hit this because we normally run Muscat in a controlled environment, but it is useful to have. Thanks!

Issue #843, opened 2019-09-03
How to create a new template?


fjorba:
I'd like to create a new template for a different document type in Muscat.  I've find a few places where I there are files that I should work on, but I think that I better ask.

Is there a guide about creating a new template?  Or could you please give me some hints?

fjorba:
Hi, I've tried to create a simple author-title-year form from scratch, but it is quite hard, and I haven't succeeded much beyond having it appear in the forms list.  Regarding the form itself, it seems that there are quite a few things that are not obvious, at least to me.

May I have a friendly hand to help me in that purpose, please?  Thanks.

I'm attaching my current version of the patch.

```
diff --git a/config/editor_profiles/default/configurations/SourceLayoutDefault.yml b/config/editor_profiles/default/configurations/SourceLayoutDefault.yml
index c8e3c37c..32729429 100644
--- a/config/editor_profiles/default/configurations/SourceLayoutDefault.yml
+++ b/config/editor_profiles/default/configurations/SourceLayoutDefault.yml
@@ -142,7 +142,34 @@ tag_exclude:
   - '773'
   - '775'
   - '787'
-
+simple_form:
+  - '026'
+  - '028'
+  - '033'
+  - '041'
+  - '100'
+  - '383'
+  - '510'
+  - '518'
+  - '546'
+  - '588'
+  - '590'
+  - '592'
+  - '594'
+  - '595'
+  - '596'
+  - '597'
+  - '598'
+  - '650'
+  - '651'
+  - '657'
+  - '690'
+  - '740'
+  - '773'
+  - '775'
+  - '787'
+  - '852'
+  
 tag_exclude_with_holdings:
   collection:
   - '506'
@@ -151,7 +178,7 @@ tag_exclude_with_holdings:
   - '591'
   - '773'
   - '852'
-
+  
 groups:
   edit_library:
     all_tags:
diff --git a/config/locales/en.yml b/config/locales/en.yml
index 59839b0f..2f901142 100644
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -244,6 +244,9 @@ en:
     collection_record_threatise_printed: "Collection record for printed treatises"
     individual_entry_threatise_printed: "Individual entry in a printed treatise collection"
     single_threatise_printed: "Single printed treatise"
+
+    simple_source: "Simple source form"
+    uab_simple_forms: "UAB simple forms"
         
   #################### MENUS
   menu_home: "Home"
[...]

PR #929, opened 2020-05-01
Explicit help for unknown tags error

fjorba:
This is my proposal to help newcomers that face the mysterious unknown_tag error.  Thanks @xhero for writing a first draft in #843.

I know my proposal only fixes the English text, but first, it needs your approval, and second, I'm still not familiarised with Gems I18N and how new texts are propagated and synced in other languages.


xhero:
I think it is a good idea! But I would place the message in the logfile, because it happens quite often in imported data that there are unconfigured tags and we could leave the full message for the developers, as users generally don't have access to the backend.

fjorba:
As a newcomer, including into the world of Rails, keeping track of everything is overwhelming. Obviously, the first thing you see and read is the screen. The current message is too terse and it does not help. As a matter of fact, without your help in #843, most likely I would have never understood what was going on.

So, if you would like to add this message to the log file, please do so. But, to help newcomers, I strongly believe that displaying on the screen is a good thing, and I'd like to keep it there, because most likely I will make errors again, and this message helps me.

xhero:
My only reserve to add a long error message is that in the (hopefully rare) case of an unknown field (it can happen in case of imported data) our users are generally not very happy to have long multi-line messages on the screen, that's why these "for the administrator" errors tend to be quite terse.
On the other hand I do understand your concern (and it is something I stumble upon quite often with new software). We could probably either add it to the docs or make a "verbose" mode for Muscat that gives more information in the error messages.

fjorba:
It depends on the use case. In my case, it is not rare, but common, as I'm taking the advantage that Muscat allows for different cataloguing templates, and the error can popup in so many different places. I took me so many hours (thanks git grep!) to find out that the 240 tag was hidden in /config/marc/#{RISM::MARC}/source/#{record_type}.marc.  For me it is a configuring and debugging aid.

Maybe this message could point to some external help document?  Something like: Please read the unknown fields in the error messages of the configuration guide?

fjorba:
I'm closing it because newer Muscat releases already display helpful messages for unknown tags (thanks!).

PR #934, opened 2020-05-11
Provide a fallback for unspecified record type

fjorba:
Records of unspecified type assume a default value of 0:

https://github.com/rism-ch/muscat/blob/develop/db/schema.rb#L442

However, this value of 0 (https://github.com/rism-ch/muscat/blob/develop/lib/marc_source.rb) is not mapped in https://github.com/rism-ch/muscat/blob/develop/config/marc/default/source/template_configuration.yml#L60.  This patch provides a reasonable source default, so they can be edited without crashing Muscat.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,9 +9,6 @@ gem 'rails', '>= 5.1', '< 5.2'
 # https://github.com/rails/rails/issues/21544
 gem 'mysql2'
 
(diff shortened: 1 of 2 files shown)


xhero:
Thanks!

fjorba:
You are welcome!

PR #938, opened 2020-05-14
Generalise the retrieval of id when saving Marc records

fjorba:
When saving a Person Marc record, or any other authority record, a popup
error appears (Error saving page!  Please try again.  error Not Found).  It
is caused because the Marc controler only considers holdings and source
records.  This patch allows saving any authority record.

### DIFF
--- a/lib/marc_controller_actions.rb
+++ b/lib/marc_controller_actions.rb
@@ -33,11 +33,11 @@ module MarcControllerActions
       # @item is used in the Marc Editor
       @item = nil
       if new_marc.get_id != "__TEMP__" 
-        # To get the ID for holdings
-        if params[:controller] == "admin/holdings"
-          @item = model.find(params[:id])
-        else
+        if params[:controller] == "admin/sources"
           @item = model.find(new_marc.get_marc_source_id)
+        else
+          # To get the ID for any other Marc record
+          @item = model.find(params[:id])
         end
       end
 


xhero:
Does this happen when saving from the marc editor? such as /admin/person/xx/edit and then saving?
That particular case was made when creating holdings from a source in the marc editor, but we regularly save people and the other authority files without crashes, so I wonder if there is another problem going on

fjorba:
Yes, when saving from the marc editor.

I suspect it happens when the marc_source field is empty. After loading external bibliographic (source) records, authority records are created without it (as temporary or minimal records), being people, institutions or standard_terms. Either directly or indirectly, this causes that the record id is not recognised and the Muscat cannot save it.

Being the message a Javascript popup, it is even a little more difficult to follow.

xhero:
Do you have an example of data that triggers this error? Or some particular configuration? I'm asking because we edit and save the authority files quite intensively, so there could be a problem in the configuration somewhere.

fjorba:
I've just confirmed what I suspected. A temporary, mostly empty authority record (created when importing external sources) fails without my patch and works with it.  For example, this is the record before and after being edited, just creating a simple main entry:

<pre>
MariaDB [traces_development]> select * from people where id=14286\G
*************************** 1. row ***************************
             id: 14286
      full_name: Lafont, Janine
    full_name_d: NULL
     life_dates: NULL
    birth_place: NULL
         gender: 0
       composer: 0
         source: NULL
alternate_names: NULL
alternate_dates: NULL
       comments: NULL
    marc_source: NULL
       wf_audit: 0
       wf_stage: 1
       wf_notes: NULL
       wf_owner: 0
     wf_version: 0
     created_at: 2019-11-26 09:06:14
     updated_at: 2019-11-26 09:06:14
   lock_version: 0
1 row in set (0.006 sec)
 
MariaDB [traces_development]> select * from people where id=14286\G
*************************** 1. row ***************************
             id: 14286
      full_name: Lafont, Janine
    full_name_d: lafont, janine
     life_dates: NULL
    birth_place: NULL
         gender: 0
       composer: 0
         source: NULL
alternate_names: 
alternate_dates: NULL
       comments: NULL
    marc_source: =100  1#$aLafont, Janine
 
       wf_audit: 0
       wf_stage: 1
       wf_notes: NULL
       wf_owner: 1
     wf_version: 0
     created_at: 2019-11-26 09:06:14
     updated_at: 2020-05-20 11:27:41
   lock_version: 1
1 row in set (0.001 sec)
</pre>

xhero:
Did you import it using the marc_import script? Normally it should disable marc scaffolding, and we generally do a second pass if it is necessary. Do you have an example of the imported marc record?

fjorba:
Yes, I have followed the Import instructions (https://github.com/rism-ch/muscat/blob/develop/3-%20IMPORT.rdoc):

$ rails runner housekeeping/import/import_from_marc.rb $FILE $MODEL

As currently we don't have proper authority records, I just imported Sources, and Muscat has been creating those minimal, temporary authority records without the marc_source or the full_name_d fields.

Regarding the exmple you requested, here is the Sources record of the equivalent Person authority record, although any one would work:

<pre>
MariaDB [traces_development]> select * from sources where id=25726\G
*************************** 1. row ***************************
          id: 25726
   source_id: NULL
 record_type: 0
   std_title: 
 std_title_d: 
    composer: 
  composer_d: 
       title: Guide de conversation français-catalan-espagnol /
     title_d: guide de conversation francais-catalan-espagnol /
  shelf_mark: [multiple copies]
    language: Unknown
   date_from: 1992
     date_to: 1992
  lib_siglum: UAB, TRACESBIB
 marc_source: =001  25726
=035  __$a025731$9TracesISISBib
=035  __$a0025-72660$9TracesVTLSBib
=245  00$aGuide de conversation français-catalan-espagnol /$cFaluba ; Morvay ; traducció francesa de Janine Lafont i de Ramon Lladó
=260  __$aBarcelona :$bLa Magrana,$c1992
=300  __$a298 p.;$c18 cm
=505  8_$aInclou índex temàtic i alfabètic
=520  3_$aInclou 300 frases d'ús habitual, 4000 mots de vocabulari i una gramàtica pràctica.
=650  04$aLèxic$0183
=650  04$aFrancès$0240
=650  04$aCatalà$0120
=650  04$aEspanyol$091
=700  1_$aLladó, Ramon$e(Trad.)$02988
=700  1_$aFaluba, Kálmán$02073
=700  1_$aMorvay, Károly$02019
=700  1_$aLafont, Janine$e(Trad.)$014286
=852  __$aUAB$bHumanitats$cRef 1 49-40$x26
=852  __$aTRACESBIB$x1

    wf_audit: 0
    wf_stage: 1
    wf_notes: NULL
    wf_owner: 1
  wf_version: 0
  created_at: 2000-09-07 00:00:00
  updated_at: 2018-02-02 11:58:04
lock_version: 0
</pre>

If not, any record that contains a unknown author will create those cases.  Just an example:

* https://ddd.uab.cat/record/221335
* https://ddd.uab.cat/record/221335/export/xm (in MarcXML, ready to import into Muscat)

fjorba:
I'm afraid this patch is still necessary, and it works, as for 6.3.1.  Without it, I cannot save any temporary authority file.

I'm pretty sure it doesn't harm your case, as the only thing it does is to generalise the case of holdings to any other record type.  It is pretty trivial and innocent.

xhero:
I'm finally catching up on this... I'm still not sure why it is necessary: if you are importing from the backend the marc controller should never be involved, that snipped is used only when you create a holding record from the source editor.

fjorba:
Ok, maybe I was misguided and there is some assumptions from Muscat that I don't know.

If I don't have authority records and I import bibliographic records, currently Muscat assigns them a correct id for each of them, so that it correctly matches authors that I'm importing.  However, the authority records lack most of the fields (see my example above https://github.com/rism-ch/muscat/pull/938#issuecomment-631418075).

Can Muscat fill automatically those values?  Does it have anything to do with scaffolding (that I don't catch)?

fjorba:
It turns out, @xhero, that you were right.  After spending some time studying this `scaffold_marc` and `supress_scaffold_marc` behaviour, I've commented out https://github.com/rism-ch/muscat/blob/91f3677d1ca37ce1ec5b1bfbce4732ada2d3b89c/lib/marc_node.rb#L258 and now the authority records' `marc_source` column gets filled and they can be edited.

I'll try better to understand this issue, maybe to pass it as parameter to `import_from_marc.rb` command line script, but now I do close this issue.

PR #940, opened 2020-05-22
Read editor specific config files that really exist

fjorba:
Creating a specific RISM::EDITOR_PROFILE profile should allow reusing
default config files, and creating only the ones that are different.  This
is what the comments in the squeeze method say, but the code didn't.  Change
it to honor this behaviour.

### DIFF
--- a/lib/editor_configuration.rb
+++ b/lib/editor_configuration.rb
@@ -39,18 +39,17 @@ class EditorConfiguration
 
     settings = Settings.new(Hash.new())
     
+    configs.each do |config|
+      default_conf = "#{Rails.root}/config/editor_profiles/default/configurations/#{config}.yml"
+      if File.exists?(default_conf)
+        settings.squeeze(Settings.new(IO.read(default_conf)))
+      end
+    end
     if RISM::EDITOR_PROFILE != ""
       configs.each do |config|
         file = "#{Rails.root}/config/editor_profiles/#{RISM::EDITOR_PROFILE}/configurations/#{config}.yml"
         if File.exists?(file)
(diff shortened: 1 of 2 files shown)


fjorba:
This is a followup from 3094d3bca6a746341db90b96caf9fe248e209d62.

I don't really know if it is better to read first the default file and then the custom one, so I just followed the comments, and I've choosed the first alternative.

Meanwhile, I'm trying to learn how to resolve merge conflicts. Please bear with me, it's my first case.

fjorba:
I understand that I cannot resolve the conflict, as I don't have write access.  It seems that only affects how the comments are reformatted. If there is anything I can do, please let me know.

fjorba:
I've just resolved the conflict updating a new version, keeping the upstream comments.

xhero:
Thanks! I'm going to test this, there is a bit of legacy stuff going on with squeeze() and I need to make sure it does not break in unexpected places. It could be a good moment for me to retouch the docs on it...

fjorba:
Great. I'm fighting with an equivalent case in lib/editor_validation.rb, but I still haven't found the solution. Damn javascript popups; tracebacks are so much more informative.

fjorba:
I've resolved a conflict with my previous commit, but I'm unsure how the final version of  lib/editor_validation.rb will be.  I have too many pending PR for me to handle with my fragile knowledge of git.

If you could please accept this pull request I'll quickly review and make sure that they are in sync.  Thanks.  Please include the others as well, including #945, because the logic of my recent patches assume it will be accepted.

fjorba:
I close this patch because I have a better solution in another one, that I'll post shortly.

PR #941, opened 2020-05-22
Make EditorValidation use get_marc_editor_profile_path

fjorba:
EditorValidation config files should use default if specific ones do not
exist, even if EDITOR_PROFILE is set to some value other than default.


fjorba:
Sorry, it seems that some of those proposals where already there.  I'm closing until I understand the conflict better.

PR #945, opened 2020-06-03
Require default EDITOR_PROFILE to be set as "default"

fjorba:
Most configuration values in config/application.rb (as copied from
config/sample_application.rb) have an explicit "default" value, except
for EDITOR_PROFILE, that may have an empty string. This complicates
the logic when this variable is used in Muscat, as it has to be
compared to empty string and then set to "default" before using it.


fjorba:
For example, in https://github.com/rism-ch/muscat/blob/develop/lib/editor_validation.rb#L111

PR #977, opened 2020-08-31
Allow edit other bibliographic records from Blacklight

fjorba:
If bibliographic records other than Sources, like Catalogues, are made
accessible through the Blacklight interface, allow them to be edited like
sources, and link them to the right record type.


fjorba:
Grazie!

PR #980, opened 2020-09-03
Document the Apache-itk option

fjorba:
Apache-itk, or the Apache multiuser-module, is a great way to simplify the
administration of multiple Apache applications in a single server.  As each
Apache runs as each user, and as the users are owned by each Apache process,
there is no longer the need to change file ownership or running sudo.
Moreover, each Apache cannot read the other users files.

As Muscat runs flawlessly under Apache-itk, document this alternative.


xhero:
Thanks!

PR #988, opened 2020-09-21
Allow $b in institution authority records

fjorba:
710 $b is already allowed in Muscat sources and holdings, but it is not
defined in institutional authority records.  Add it as optional.  Change
also $a to be required and not repeteable in 110, 410 and 510, if the tag is
defined.


xhero:
I need to ask my colleagues in ZR about this: @HirschSt @jenniferward would it work for us?

jenniferward:
It's fine with me, but maybe Guido can confirm?

Docudoctor:
I confirm. 
Guido

xhero:
Super! Thanks to everybody, I will merge this.

Issue #996, opened 2020-10-01
Problems with SourceLabel translations


ahankinson:
There are a few errors and issues with `config/editor_profiles/default/configurations/SourceLabels.yml`. 

1. There are quite a few duplicated language keys, with different values. I'm guessing this is a copy/paste error, where someone copied an existing entry and changed the translation, but did not change the key. For example: https://github.com/rism-ch/muscat/blob/develop/config/editor_profiles/default/configurations/SourceLabels.yml#L78-L85, where there are two `it` translations, and one is probably supposed to be `es`. 
2. Having all the MARC indicator codes in the same level means there are duplicate keys. For example, `art` is used for both "Artist" (as a relator code) and "Artificial" (as a language). See https://github.com/rism-ch/muscat/blob/develop/config/editor_profiles/default/configurations/SourceLabels.yml#L2881 and https://github.com/rism-ch/muscat/blob/develop/config/editor_profiles/default/configurations/SourceLabels.yml#L3372
3. There are duplicate entries for the same codes. For example, `asn` is defined twice, slightly differently: https://github.com/rism-ch/muscat/blob/develop/config/editor_profiles/default/configurations/SourceLabels.yml#L2898, https://github.com/rism-ch/muscat/blob/develop/config/editor_profiles/default/configurations/SourceLabels.yml#L3078. This is also a problem with `ccp`. 

I have gone through and fixed the issues with 1 where I could and have filed a PR for review. Issue 3 will require someone to go through and select the correct translation, and remove the other one. For issue 2 it would benefit from further subdividing the configuration into sections, like `relators:`, `languages:`, etc. to avoid the naming collisions.

ahankinson:
Most of the other label configurations have similar problems. (My code editor highlights the lines with errors, so it may be worth going through with a similar tool to check it. Interestingly, many YAML linting tools do not catch this, but this one does:

https://jsonformatter.org/yaml-validator

ahankinson:
Additionally, some of the keys are in uppercase, e.g. `IT` instead of `it`.

lpugin:
> For issue 2 it would benefit from further subdividing the configuration into sections, like relators:, languages:, etc. to avoid the naming collisions

Maybe another option could be to prefix them. `lang.art` vs `relator.art`?

ahankinson:
In YAML the hierarchy structure parses out to a map / dict / JSON object, so that’s effectively what it’s doing anyway.

ahankinson:
[...]

PR #997, opened 2020-10-01
Fixed: conflicting language keys in SourceLabels

ahankinson:
Refs #996

### DIFF
--- a/config/editor_profiles/default/configurations/SourceLabels.yml
+++ b/config/editor_profiles/default/configurations/SourceLabels.yml
@@ -81,7 +81,7 @@ doc_abbr_voices_instr:
     en: "Terms for voices and instruments"
     it: "Terminologia per voci e strumenti"
     pt: Termos para vozes e instrumentos
-    it: "Terminología para voces e instrumentos"
+    es: "Terminología para voces e instrumentos"
     pl: Określenia głosów i instrumentów
 doc_abbr_voices_instr2:
   label:
@@ -89,7 +89,7 @@ doc_abbr_voices_instr2:
     en: "RISM instrument abbreviations"
     it: "Abbreviazioni RISM per gli strumenti"
     pt: "Abreviaturas de instrumentos do RISM"
-    pt: "Abreviaturas de instrumentos de RISM"
+    es: "Abreviaturas de instrumentos de RISM"
     pl: Skróty instrumentów wg RISM
 
 ###############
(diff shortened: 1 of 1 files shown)


config/editor_profiles/default/configurations/SourceLabels.yml
```diff
@@ -3980,15 +3971,6 @@ smn:
...
     pl: Samski inarski
-srp:
-  label:
-    IT: Serbo
-    FR: Serbe
-    de: Serbisch
-    en: Serbian
-    pt: "Sérvio"
-    es: Serbio
-    pl: Serbski
```
lpugin, line 3991: Is this one a duplicate?

config/editor_profiles/default/configurations/SourceLabels.yml
```diff
@@ -2895,15 +2895,6 @@ asg:
...
     pl: Cesjonariusz
-asn:
-  label:
-    it: Altro nome
-    fr: Autre nom
-    de: Weiterer Name
-    en: Associated name
-    pt: "Nome associado [asn]"
-    es: Nombre asociado
-    pl: Powiązana osoba
```
lpugin, line 2906: @jenniferward could you double check this? (The key is duplicated but with different values)

config/editor_profiles/default/configurations/SourceLabels.yml
ahankinson: Yes:

https://github.com/rism-ch/muscat/blob/develop/config/editor_profiles/default/configurations/SourceLabels.yml#L3983

https://github.com/rism-ch/muscat/blob/develop/config/editor_profiles/default/configurations/SourceLabels.yml#L4019

config/editor_profiles/default/configurations/SourceLabels.yml
jenniferward: I don't see where this is used in Muscat; I can't find  "Weiterer Name" on any of the 700 relator code lists.

lpugin: @xhero could you check if this is used at all? I think it would be best to remove from the translation list all the relator codes that are not used.

ahankinson: I would suggest punting the cleanup of the relator codes until after we've transitioned to the consolidated translations. It will probably be easier then.

xhero:
I'm going to merge this, the language codes will be fixed again anyways

Issue #1013, opened 2020-10-26
Should secondary literature have more than one template?


fjorba:
Currently, in Muscat there are different templates for sources, because different document types require different fields (Marc tags and subfields) and defaults.  I think that secondary literature should have the same capability, because the document types are so different that they require different fields (for example: a journal article and a book monograph).

I volunteer to prepare the infrastructure for that, now that the two models are aligned (#987).  My proposal is to start with a list of a single document type, the current one (we can call it Default, to start with), but that would open the door to easily create others.

What do you think?

MFalletta:
Great idea! I agree

jenniferward:
Me too!
If it helps, I made a very short guide (aimed at Muscat contributors) that talks about the different types of secondary literature in Muscat. It briefly explains fields needed for different types of materials - it might give you some ideas:
http://www.rism.info/fileadmin/content/community-content/Muscat_EN/Secondary_lit_brief_guide.pdf

fjorba:
Thank you, I'll take it into account.

xhero:
I think it is doable, you would add a record_type like in source right?

fjorba:
My idea was to mimick source model as much as possible, with the obvious changes (composer to author, for example). 

> I think it is doable, you would add a record_type like in source right?

I have a prototype working in my local machine, but now I don't remember the details, it was just a proof of concept that I could do it.  It was running some weeks ago, I'll have to refresh my memory.  But yes, if you count on record_type being there, I'll double check that it is there.

Obviously I'll require your review, and I'll appreciate your comments.

ahankinson:
A heads-up that there are some changes coming for languages (#1004) that will probably affect this work, since (I believe) you will need a "Labels" file that gives the translations for each field.

fjorba:
What a nice moment for such a change!  (Yes, I'm sure it is never a good moment for such a change...)

Do you consider that I should your patch before I publish mine, so I can understand how it affects the rest of the cataloguing files?  My idea is to prepare the infrastructure for choosing templates for 7.0.  If I can create a new one (maybe starting with journal article), great, but that would be a second step.

xhero:
I think at the end it should not really change that much on the implementation side of things for the templates, as the languages are managed (more or less) transparently in the backend.
[...]

Issue #1004, opened 2020-10-09
Move translations to standard Rails i18n mechanism


ahankinson:
This is an issue to track the progress of a translation clean-up that is underway and ensure visibility of the process across the team. 

Currently the editor uses Muscat-specific format for configuring the field names, where each type of record (Source, Person, Work, etc.) has a complete set of translations for all languages. This has led to divergent translations, and a lot of work for translators as they need to go through each file and translate each key.

The goal of this mini-project is to shift the editor field translations to a more standard Rails i18n configuration mechanism, and to consolidate the translations so that we can better ensure consistency across languages. This means there will  be a central set of translations in the `locales` directory, with a mapping from the MARC editor configurations to a shared unique key in these translations.

For example, we currently have:

```
"246":
  label:
    it: Varianti del titolo diplomatico
    fr: Variante du titre diplomatique
    de: Weiterer diplomatischer Titel
    en: Variant title on source
    pt: "Título variante na fonte"
    es: "Variante de título en fuente"
    pl: Alternatywny tytuł w źródle
  fields:
    a:
      label:
        it: Varianti del titolo diplomatico
        fr: Variante du titre diplomatique
        de: Weiterer diplomatischer Titel
        en: Variant title on manuscript
        pt: "Variante de título na fonte"
        es: "Variante de título en fuente"
        pl: Alternatywny tytuł w rękopisie
      edit_label:
        en: ""
```

This will be centralized and cleaned up (in a spreadsheet), assigning a unique key to each translation:

```
records.variant_ms_title | Variant title on manuscript | Variante du titre diplomatique | Weiterer diplomatischer Titel | Varianti del titolo diplomatico | Variante de título en fuente | Variante de título na fonte | Alternatywny tytuł w rękopisie
---
records.variant_source_title | Variant title on source | Variante du titre diplomatique | Weiterer diplomatischer Titel | Varianti del titolo diplomatico | Variante de título en fuente | Título variante na fonte | Alternatywny tytuł w źródle
```

The mapping will then look like this:

```
"246":
  "fields":
    "a":
      "label": "records.variant_ms_title"
  "label": "records.variant_source_title"
```

This will expand to the value of the "records.variant_ms_title" key in separate translation yml files, e.g., for German a file "de.yml" would have:

```
"de":
  ...
  "records":
      ...
      variant_ms_title: Weiterer diplomatischer Titel
      variant_source_title: Weiterer diplomatischer Titel
```

While "fr.yml" would have:

```
[...]

PR #1017, opened 2020-11-02
Allow different templates for secondary literature

fjorba:
This patch is the first step towards having different templates for
secondary literature.  It synchronises some code and configuration files
from sources so now, when creating a new record, appears a menu with a
single choice, the legacy template.

There are yet no new fields (like record type) or tag changes.

### DIFF
--- a/app/admin/catalogue.rb
+++ b/app/admin/catalogue.rb
@@ -1,14 +1,14 @@
 ActiveAdmin.register Catalogue do
-  
+
   include MergeControllerActions
-  
+
   collection_action :autocomplete_catalogue_name, :method => :get
 
   menu :parent => "indexes_menu", :label => proc {I18n.t(:menu_catalogues)}
 
   # Remove mass-delete action
   batch_action :destroy, false
-  
+
   # Remove all action items
   config.clear_action_items!
   config.per_page = [10, 30, 50, 100]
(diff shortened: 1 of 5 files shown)


fjorba:
This patch does not close, but is a first step towards #1013.

@xhero, sorry but I don't know how to deal with the translations of a couple of new messages I've created in en.yml. Do I have to create the entries in the other languages yaml files? In English? Moreover, with #1004 in the works, I still have more doubts.

I've tested this patch it in my three different local Muscat configurations, but please review it. I'm unable to find a way to request a proper review in Github.  It seems that it is due because I don't have enough privileges (I'm not part of the organization or I'm not Pro).

xhero:
Thanks! I will review it a bit later today or tomorrow morning, at a first glance it looks ok!

xhero:
It seems all ok on my side - I merged it in its own branch. Thanks!

xhero:
I forgot to add: for the translations, for the moment simply add them to the other languages in capital letters if you cannot translate them, i.e.:
(in it.yml)
catalogue: CATALOGUE

this way we can spot them and translate them. Regarding #1004 I think we can backport them by hand, unless there are many many translations to add.

fjorba:
Thanks, I'll add them as soon as I can. I'll create another patch; it's just a couple of texts, that hopefully will become obsolete some time in the future.

PR #1019, opened 2020-11-05
Fixed: Default to warning in production

ahankinson:
This changes the default log level for production environments to WARN from DEBUG. This helps keep the file sizes of logs in production down.


xhero:
Thanks :)

PR #1020, opened 2020-11-06
Fix default blacklight Solr settings

ahankinson:
The Blacklight settings point to 'default' being the default core for Blacklight in production, but sunspot configures the default core to be 'production'. 

This PR aligns the two by changing the default setting for Blacklight to 'production'.


xhero (approved):
This is a weird bug in Blacklight, because sometimes it wants "default" and sometimes  "production" and it changes between restarts (!). I think it is caused by sunspot that aliases "default" with "production" at the filesystem level. I spent lots of time on this but could never find what caused it. Hopefully when we go with solr 8.6 and properly configure it it will not happen again. Having production or default is harmless since sometimes it will have to be changes between apache restarts.

PR #1021, opened 2020-11-08
Document sample_application.rb variables

fjorba:
Make sample_application.rb friendlier to newcomers by documenting each
variable.  Reorder slightly some of them to put the most important first,
remove obsolete indentation, normalise spaces and quote signs, and make the
file utf-8 friendly.

### DIFF
--- a/config/sample_application.rb
+++ b/config/sample_application.rb
@@ -1,120 +1,155 @@
-require File.expand_path('../boot', __FILE__)
+# coding: utf-8
 
-require 'rails/all'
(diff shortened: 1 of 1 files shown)


fjorba:
By the way, while searching the usage of those variables, I found some obsolete code and maybe some of them may have another, more meaningful name, according to its current usage.  But I'll open an issue and a pull request for that.  In this case, as there is no real change, I thought that a plain pull request was enough.

xhero:
Thanks! yes there is some unused stuff, I will be making some cleanups too

Issue #1028, opened 2020-11-18
Secondary literature table rename: migration tasks


fjorba:
There are a few db migration tasks for the secondary literature table rename.  As they are scattered among different issues, I think that it is better to document them in a single one, so @xhero can find it easily:

Table "catalogues" (rename to: "publications", see #975)

Columns:

t.string "name"  (_rename to_: "short_name", see #1027)
               (_new integer field_ -> "record_type", see #1013)
t.string "author"
               (_new string field_ -> "author_d", to match sources table)
t.string "description" (_rename to_: "title", see #626)
               (_new string field_ -> "title_d", to match sources table)
t.string "revue_title" (_rename to_: "journal", see #626)
               (_new string field_ -> "journal_d", to match sources table)
t.string "volume"
               (_new string field_) -> "number" (not discussed yet anywhere, but it is needed)
t.string "place"
t.string "date"
t.string "pages"
t.integer "wf_audit", default: 0
t.integer "wf_stage", default: 0
t.string "wf_notes"
t.integer "wf_owner", default: 0
t.integer "wf_version", default: 0
t.integer "src_count", default: 0
t.datetime "created_at"
t.datetime "updated_at"
t.text "marc_source"     (change type to: mediumtext, see #792)
t.integer "lock_version", default: 0, null: false
t.index ["name"], name: "index_catalogues_on_name"
t.index ["wf_stage"], name: "index_catalogues_on_wf_stage"

xhero:
I did the first step! The others (like renaming) will need a bit of testing so I would like to do them as a next step. We are planning a release next week, I think we can add all these mods just after (we plan another one in a month or so)

fjorba:
Of course; please let me know which steps am I supposed to do, because I cannot easily distinguish them.  I think that db migration changes should be done by you, @xhero, but if you consider that there is something  in the software that I should take care of, please let me know and I'll do my best.

PR #1043, opened 2020-12-15
Allow different templates for secondary literature

fjorba:
This patch is the second version of the first step towards having different
templates for secondary literature.  It synchronises some code and
configuration files from sources so now, when creating a new record, appears
a menu with a single choice, the legacy template.

This version takes into account the catalogues to publications rename and it
adds the locale entries compatible with the i18n reorganisation.

There are yet no new fields (like record type) or tag changes.


fjorba:
This patch updates and replaces #1017.

config/locales/de.yml
```diff
@@ -246,6 +246,9 @@ de:
     individual_entry_threatise_printed: "Theoreticum, gedruckt - Teileintrag der Sammlung"
     single_threatise_printed: "Theoreticum, gedruckt - Individualeintrag"
   
+    legacy_templates: "LEGACY TEMPLATES"
```
ahankinson, line 249: The locales are updated upstream in https://github.com/rism-digital/translations. Any changes here will be overwritten when we next sync them up.

Also, we are trying to not put "placeholders" in the language files, so every new entry should be translated into its appropriate language before being added.

app/admin/publication.rb
```diff
@@ -1,14 +1,14 @@
 ActiveAdmin.register Publication do
-
```
ahankinson, line 2: Having these changes in this patch makes it difficult to review this change.

config/locales/de.yml
fjorba: I just followed @xhero advice: https://github.com/rism-ch/muscat/pull/1017#issuecomment-721639815

So, which is the preferred way now?

app/admin/publication.rb
fjorba: I understand, but Emacs strips those extra spaces for me.  In fact, there are so many in Muscat that it is difficult for me to stop Emacs to strip them everywhere else.  Shouldn't be a policy in Muscat, like in other projects, to strip them before commiting them?  I know that there are git hooks for that.

fjorba: On the other hand, I have just seen that from the Github web interface it is possible to hide whitespace changes, using the icon in the middle of the **Review changes**  button line.

fjorba:
May I do anything to help the approval of this path?

xhero:
Sorry! I'm a bit late on this track, I will follow a review shortly!

xhero:
Everything in the patch seems ok to me, for the moment we can keep then placeholder since they are in the "interface" language file.
I only have one question: where do you store the resulting record_type? In Source there is a dedicated field, but I can't find one here.

fjorba:
> I only have one question: where do you store the resulting record_type? In Source there is a dedicated field, but I can't find one here.

It's requested in #1028 (I've also put some other requests on hold, because maybe there are better solutions).

xhero:
ok, for the moment I would add a record_type as you suggest, I can also rename title to description

xhero:
Sorry for the silence - I had to take a small detour for other projects, I will add those fields in 7.1 (and in the next dev branch) so we can merge this.

fjorba:
Can I do anything else for this (and my other pending PR) to be included upstream,  @xhero?  I'm applying those  patches to my sites and we are using them daily (although not in production yet). Thanks.

fjorba:
Obsolete, I'll publish a rewrite later on.

PR #1050, opened 2020-12-31
Remove obsolete UrlHelper

fjorba:
Current behaviour of UrlHelper creates a innecessary redirect from /sources
to /catalog, probably for some old requirement no longer valid.  Remove
local UrlHelper and inherit generic Blacklight behaviour (by default,
/catalog) and the second part of those redirects in routes.rb.

### DIFF
--- a/app/helpers/url_helper.rb
+++ b/app/helpers/url_helper.rb
@@ -1,8 +1,0 @@
-module UrlHelper
-  include Blacklight::UrlHelperBehavior
-
-  def url_for_document(doc, options = {})
-    return "" if !doc
-    "/" + params[:controller] + "/" + doc.to_param
-  end
-end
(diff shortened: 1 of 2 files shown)


fjorba:
Tested both on a sources-based local Muscat instance and on a secondary-literature one.

xhero:
This is ok for me, but I need to do a bit of testing since we have some users that still have permalinks with the old /manuscripts/ url and may need them.

fjorba:
I understand; maybe you could just remove the url_helper.rb file but keep the manuscripts routes.rb. But then, a comment to say that is for backwards compatibility for older urls would help to understand why it is there.

I found this file when trying to change /catalog to /record for compatibility with our current systems.  Now I know how to do it (I have to modify just 3 lines!), and playing around I found that this file was creating confusion rather than clarifying things.

fjorba:
Closing because somehow it was included in ee91ba9d009ecaff35fe3bc8c998bf2539ab3ded, although it is not shown in this PR.

PR #1051, opened 2021-01-05
Reorder publication.rb model like source.rb

fjorba:
Publication model shares with Source that both are bibliographic Marc
records.  Reorder the common lines of both models so that the common ones
are aligned and is easier to spot the differences, as already done for the
app/admin/*.rb files.

The only additions in publication.rb are a few comments when they make sense
and make the differences smaller.

Strip trailing blanks in both files so, again, the comparison between them
is cleaner.


fjorba:
This patch is part of a series of gradually improving secondary literature (publications), with small, conservative changes. As each depend on the previous one, I'd sincerely appreciate your approval without much delay. Thanks.

xhero:
Thanks! It seems all ok to me!

PR #1052, opened 2021-01-06
Introducing SITE_* configuration variables

fjorba:
Rename a few configuration variables with a SITE_ prefix, and change them
from PROJECT to TITLE, so they have a less provisional name.  Of special
importance is the old RISM::BASE, now promoted to RISM::SITE_ID and put at
the first one of the application.rb file, so it can be used already in this
file even to conditionally define the rest of them.  Clarify their use when
defining logos, so newcomers will spend less time guessing and learning via
trial and error.


xhero:
Thanks, I think it is all ok and SITE_* works for me.

PR #1058, opened 2021-01-13
Index :wf_stage in Publications like Source

fjorba:
:wf_stage in Solr allows Muscat Blacklight interface search by publications
approved or not, so they can be restricted to public visibility, like
Sources.  Add it to paper_trail as well.  Reorder a few more lines to match
Sources.


xhero:
Thanks!

PR #1059, opened 2021-01-13
Fix edit link for sources in faceted search

### LINKED ISSUE #1057: Faceted search links to "Publications"
I think this is a bug: When you are in the Faceted Search and looking at a record, the "Edit" button in the upper right corner leads to the Publications (Secondary Literature), and of course the record is not found.

Example: https://muscat.rism.info/catalog/1001058746, Edit, which links to https://muscat.rism.info/admin/publications/1001058746/edit

fjorba:
Record type is in singular (Source) not plural (Sources). Fixes #1057


xhero:
Thanks! I'm a bit late with the reviews, I will be catching up shortly!

PR #1064, opened 2021-01-15
Fixed: Improved output of the IIIF manifest script

ahankinson:
Some improvements to the manifest generation script for Muscat. Note that this hasn't actually been tested, since I don't have a working Muscat installation locally.

 - Changes the path to the new RISM Digital IIIF server
 - Fixes the canvas ID property to be a full URI
 - Improves the 'related' property with a label and format per the spec recommendations
 - Improves the human readable label of the canvas to an enumerated image value instead of the filename
 - Updates the value written back to Muscat


ahankinson:
Before this is merged, we should probably talk about the format for IIIF image URIs, since we will need to figure out how they are stored on the image server.

PR #1076, opened 2021-03-10
7.0 devel et

HirschSt:


xhero:
The PR looks good to me, I'm just wondering if we should have a localised version of "standard", "additional" and "text"?

HirschSt:
Thanks. Yes, but since this column is only visible to editors and admins I think we can postpone this for later improvement

fjorba:
I'm still seeing a few references to old name _catalogues_ instead of current _publications_ in this patch, specifically in the schema.rb (ex: https://github.com/rism-digital/muscat/pull/1076/commits/eb04370a6fc98002c9232b521ae2bd318cbf2f7e#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3) , routes.rb files, but also in views. Am I reading it wrong, or still there are some leftovers?

xhero:
Yes well spotted! It was a left-over table, I made a migration to suppress it.

PR #1094, opened 2021-04-10
Change local variable name composer to full_name

fjorba:
The MarcPerson class is used both for sources and publications.  The
composer and composer_d variables are certainly local, and this change is
mostly useless, except that it now matches the function name, clarifies its
use and helps contributors familiarisating with Muscat code that it generic,
not only for musical sources.


xhero:
Thanks, well spotted!

PR #1097, opened 2021-04-29
Change 'Cataloghi tematici' to 'Bibliografia'

baccia:


ahankinson:
This should actually be changed in rism-digital/translations so that we can have the same translations in Muscat and Muscat Plus.

ahankinson:
Ah, actually -- nevermind. The file you're changing is not in that repo.

PR #1101, opened 2021-05-11
Fix installation issues

tramuntanal:
While installing Muscat I've found some issues that should be manually resolved in order to get the application up and running. Some are specifically related to the test environment, and some are common to all environments. This PR tries to fix the issues.

In this PR:
- [x] The migration `db/migrate/20140401072119_create_catalogues.rb` crashes
-  [ ] In the test environment the ports where Solr is listening on and the port used to connect are different.
- [x] There are no seeds for all kinds of users. Having a user for each role will facilitate development.


tramuntanal:
I see there's `db/test.sqlite3` checked into version control. What's the point on having this file checked?

xhero (commented):
Is this because of the Blacklight configuration (https://github.com/rism-digital/muscat/blob/develop/config/blacklight.yml#L16-L18)?
I would probably change 8982 there and leave here the default for now.

fjorba:
I think that the migration issue (db/migrate/20140401072119_create_catalogues.rb) is related to the :catalogues to :publications rename (1fb8c04275c304a6a1243fa2347d58951dcfee3c). It refers to an older schema. Probably the fix is fine, but now it is fixed because the value already changed in https://github.com/rism-digital/muscat/blob/develop/config/sample_application.rb

xhero:
Yes good catch - @tramuntanal  probably we could amend the migration to use :publication from the configuration file? So in the future if it is needed we can initialise a DB with non zero ids.

tramuntanal:
Hi @xhero I've applied your recommendations and also have removed the file `db/test.sqlite3` and add it to `.gitignore`.

I've also fixed the :user factory and improved tests from 98 failing to 13 failures. Test suite looks different now.

Moving this PR to from draft to ready.

xhero:
looks good, thanks!

PR #1102, opened 2021-05-14
Add record_type do publications

fjorba:
Add the same field as in sources, as it will be needed as we will create
different templates.  Please note that:

1.  At this moment there is yet no use for this field.  This will come
later.
2.  I haven't attached the db/schema.rb file in this patch due to a
conflict.  This is my first migration and I'm unsure how to proceed, but
fortunately db/schema.rb is (re)generated automatically after performing
'rails db:migrate', so please do this step for me.  Thanks.


fjorba:
Part of #1028

PR #1103, opened 2021-05-14
Create indexes for created_at and updated_at in publications

fjorba:
Create the same indexes as in sources, as it is needed for fast date
searching.

Please note that I haven't attached the db/schema.rb file in this patch due
to a conflict and and I'm unsure how to proceed, but fortunately
db/schema.rb is (re)generated automatically after performing 'rails
db:migrate', so please do this step for me.  Thanks.


xhero:
OK, I will add the new schema after I do the migration.

Issue #1027, opened 2020-11-18
Proposal: rename the "name" column to "short_name" in the Secondary literature table


fjorba:
(I'm preparing an issue that gathers and links to all name and db migrations for the Secondary literature Catalogues to Publication change, see #975.  However, I think that each one should be discussed separately, to reach consensus.)

In the current "catalogues" (secondary literature) table there is a column (field) named "name":
https://github.com/rism-ch/muscat/blob/02aca2baf1d5e169b2f1fbbf29326a1a107d1c75/db/schema.rb#L41

The current usage of this field is to univocally identify a secondary literature publication with a short name or identifier. Currently, RISM extracts this value from Marc 210 $a (https://www.loc.gov/marc/bibliographic/bd210.html).  Some RISM examples:

```
210 __ $a Sacriconcentus
210 10 $a DTB 27-28
210 __ $a TerP
210 10 $a SzkL
210 __ $a BaranyaiHelytörténetirás
210 __ $a RennoP 2018
210 __ $a IlonaM 1989
210 __ $a FogD 1984
210 __ $a LesureP 1975
210 __ $a FogN 1976
```

I think that "name" doesn't match its real use, as in the secondary literature cataloguing template it is shown as "Short title (210)", and in the Active Record label, also "Short title" (https://github.com/rism-ch/muscat/blob/02aca2baf1d5e169b2f1fbbf29326a1a107d1c75/app/admin/catalogue.rb#L154).

My proposal to change it to "short_name" is double: first, that "name" it is not a visible name, because the visible one is "Short title" (introduced in 26661d4c3bf6aedc0721dd23a558dd5a0f3bc791 and 9d7dc1b83f39ce97a1d3b0c519b4ddd25f6ea86a).  Second, to have a less ambiguous name: searching for "name" in Muscat gives too many false positives, whilst "short_name" currently is not used anywhere (even this sentence is confusing due to that!).

[...]

Issue #626, opened 2018-04-13
Secondary literature: change 240 to 245


jenniferward:
In the template for a new bibliographic item for Secondary literature:
http://muscat.rism.info/admin/catalogues/new
the field number 240 is incorrect. In this field, "Title" is actually the title as given on the title page, so the field should be designated 245 instead. At the moment we do not need a separate field for 240.

HirschSt:
needs maintenance

HirschSt:
check if all 240 subfields are in 245 configuration

jenniferward:
We only need a 240 when we catalog music editions (which need uniform/standardized titles), but these do not occur very often, and I think having both would confuse many users. So far it looks like people have input music editions in 245 style anyways (i.e. statement of title as on title page). And if we're going to use standardized titles, then we need the full range of key/opus/catalog of works options, which I think we want to avoid.

I might suggest that the current 240 $g gets changed/migrated to 245 $k and 240 $h gets changed to 245 $h (if we want to keep both of those in the 245).

I think the information in the 730 is better suited to a 246 field.

lpugin:
@HirschSt will look when this can be updated on the OPAC side. Once this is clarified we can change it and update the data.

fjorba:
At the same time, I'd say that, if 245 is the tag chosen for title, the schema could change the field name `description` for `title`, because title is it:

https://github.com/rism-ch/muscat/blob/d272db5b35f67b23d349101d8bccfb3a9b966dee/db/schema.rb#L43

And, as a wish-list, `revue_title` for `journal`, as it is most commonly used in academic literature:

https://github.com/rism-ch/muscat/blob/d272db5b35f67b23d349101d8bccfb3a9b966dee/db/schema.rb#L44

jenniferward:
Yes to "journal"!

fjorba:
I think I can take care of this one, just after pushing #1104. However, I don't know how does it affect your current database records, that I think you can solve better than me.

fjorba:
I have done the easy part, the trivial search-and-replace. Now, I need your input, @jenniferward or whoever else decides, about what to do with the old 240 $g.  In principle, it seems that 245 $h is the closest equivalent, albeit there are differences:

* https://www.loc.gov/marc/bibliographic/bd240.html
* https://www.loc.gov/marc/bibliographic/bd245.html

[...]

PR #1104, opened 2021-05-16
Rename description to title in publications

fjorba:
As part of the publications reorganisation, change the old name description
to title. Perform the database migration and change all occurences.

Please note that in this commit there is no db/schema.rb due to a local
mismatch.  Please add in after performing 'rails db:migrate'.  Thanks.


fjorba:
Part of #1028, and #626

xhero:
It looks ok to me, we need to do a bit of testing to make sure all the references are caught, I propose to do this in 7.2 so we are not pressed, since 7.1 should come quite soon.

fjorba:
Understood.  While working on this, I've come across some name variables and methods not yet changed to short_name (#1027). Is this change going to be included in 7.1? Because I think I'd better fix it in the same release.

fjorba:
Hello, @xhero, is there a chance to push this patch? I'm experiencing some errors regarding a mismatch between title and description in the Publications model, and my guess is that it will be easier if we complete this transition, as now it is difficult to know what is done and what is left.

Thanks.

xhero:
I merged this patch but I would prefer to avoid patches like this one in the future unless these is a pressing need for an important reason to rename a field on the database, since a modification like this touches many different parts of the system and can lead to some unexpected behaivour which is difficult to test. At the same time, it bings no real improvement to the system, as the db field name is used only internally.

fjorba:
I understand, and it makes sense.  I'll take it into account, thanks for the explanation.

Issue #991, opened 2020-09-28
Add a brief_record column for Sources and Secondary Literature tables


fjorba:
Brief records in Muscat currently are very brief indeed, as they only show the composer and the title.  Currently they rely on the fields Solr returns, provided they are indexed with the stored property.  This brief information is often not enough to differenciate between records.  But adding fields to Solr, moreover with this stored property, just to have them displayed in the brief record may not be efficient.

Let me display with and example the difference between a simple author-title and richer records, searching the same author:

* http://www.rism-ch.org/catalog?utf8=%E2%9C%93&locale=en&search_field=composer&q=palestrina
* https://ifmuc.uab.cat/search?sc=1&p=palestrina&f=author&ln=en

A normal user, searching a catalog-like site, forces the display of dozens or hundreds of brief records, so they may be displayed fast, with a minimum of computing overhead.  Most of the systems I've worked with have the brief record pre-formated in a table, so they are just retrieved from there and send to the screen as-is.  With this strategy, in the faceted search (Blacklight) only the record id is needed from Solr, and this id will be used to retrieve the brief record from the database table. (Active record in the internal Muscat work area is not the target of this issue.)

Details of how and when this field is created will be defined in a future issue, but let me advance here some of the basic ideas I have in mind.

Which fields are chosen to this play brief records, and how they are formatted will probably different from site to site, they may change depending on the record type, due to personal criteria, changing opinions or requirements.  In my experience, creating them may require some sophisticatcation, because the how some fields are shown may depend on the existence of another one, or even the number of yet another field.  My experience is that templates are never enough, and they must be completed with a real programming language.  So, I'd leave them to be written in Ruby; I would't even try to create a template.  There should be a default implementation upstream, but may be modified in each instance.  I volunteer to write a first version.

The brief record column should be updated whenever the record is updated, but there should be a command line utility to update them (probably in record subsets) if the brief record definition has changes.

As hopefully there is a database migration in the horizon (#975), I think that it would be a great moment to add this column in both bibliographic records models.  Text type would be enough; medium text is not needed here.

xhero:
[...]

Issue #958, opened 2020-07-07
Secondary literature (Catalogues) not searchable via the Faceted search in Muscat 6.2


fjorba:
I've downloaded all RISM catalog files from https://opac.rism.info/main-menu-/kachelmenu/data, extracted catalogues.xml, and uploaded it to a clean Muscat 6.2 instance, using the instructions from https://github.com/rism-ch/muscat/blob/develop/3-%20IMPORT.rdoc:

<pre>
$ rails runner housekeeping/import/import_from_marc.rb catalogues.xml Catalogue
</pre>

Import went right, and tables have been populated accordingly (those folders items are tests of mine, just to check that the functionality was working fine):

<pre>
MariaDB [muscat_development]> select table_name, table_rows from information_schema.tables where table_schema = 'muscat_development' order by table_name;

+--------------------------------+------------+
| table_name                     | table_rows |
+--------------------------------+------------+
| active_admin_comments          |          0 |
| ar_internal_metadata           |          1 |
| bookmarks                      |          0 |
| catalogues                     |      42814 |
| catalogues_catalogues          |          0 |
| catalogues_to_catalogues       |       6038 |
| catalogues_to_institutions     |       2603 |
| catalogues_to_people           |      19783 |
| catalogues_to_places           |       2261 |
| catalogues_to_standard_terms   |       2175 |
| crono_jobs                     |          0 |
| delayed_jobs                   |          0 |
| digital_objects                |          0 |
| digital_object_links           |          0 |
| folders                        |          2 |
| folder_items                   |         81 |
| holdings                       |          0 |
| holdings_institutions          |          0 |
| holdings_to_catalogues         |          0 |
| holdings_to_people             |          0 |
| holdings_to_places             |          0 |
| institutions                   |       1345 |
| institutions_institutions      |          0 |
| institutions_to_catalogues     |          0 |
| institutions_to_people         |          0 |
| institutions_to_places         |          0 |
| institutions_to_standard_terms |          0 |
| institutions_users             |          0 |
| institutions_workgroups        |          0 |
| liturgical_feasts              |          0 |
| people                         |       9535 |
| people_to_catalogues           |          0 |
| people_to_institutions         |          0 |
| people_to_people               |          0 |
| people_to_places               |          0 |
| places                         |        724 |
| roles                          |          5 |
| schema_migrations              |         79 |
| searches                       |          6 |
| sources                        |          0 |
| sources_to_catalogues          |          0 |
| sources_to_institutions        |          0 |
[...]

PR #1106, opened 2021-05-18
New global config variable: MAIN_BIBLIOGRAPHIC_RECORD_TYPE

fjorba:
Muscat was created for cataloguing musical manuscripts (Sources), but it
also allows cataloguing secondary literature (Publications).  There is some
work going on to adapt Muscat to be used as institutional or research
repository, where the main bibliographic record type would be Publications.
This first patch allows to define this parameter, and just corrects the
public, Blacklight Solr interface.

### DIFF
--- a/app/views/catalog/_show_default.html.erb
+++ b/app/views/catalog/_show_default.html.erb
@@ -4,12 +4,17 @@
   This is pretty hackish for now - figure out a better way to do this into the CatalogController! -%>
 
 <%
-	@item = Source.find(@document.id.split(" ")[1]) # hackady-hack! Source 1234567 -> 1234567
+	if RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE == "Source"
+	  # hackady-hack! Source 1234567 -> 1234567
+	  @item = Source.find(@document.id.split[1])
+	else
+	  @item = Publication.find(@document.id.split[1])
+	end
 	@editor_profile = EditorConfiguration.get_show_layout @item
 %>
 <dl class="dl-horizontal  dl-invert">
 	<%= render :partial => "marc/show", :locals => { :opac => true } %>
-	<%- if @item.digital_objects.size > 0 %>
+	<%- if @item.has_attribute?(:digital_objects) && @item.digital_objects.size > 0 %>
 		<%= render :partial => "catalog/digital_objects" %>
 	<%- end %>
 </dl>
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -179,7 +179,7 @@ class CatalogController < ApplicationController
       :"q.alt" => "*:*",
       :rows => 20,
       :defType => 'edismax',
-      :fq => "+type:#{RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE} wf_stage_s:published",
+      :fq => "+type:#{RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE} +wf_stage_s:published",
       :hl => 'false',
       :"hl.simple.pre" => '<span class="highlight">',
       :"hl.simple.post" => "</span>",
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -118,7 +118,7 @@ class CatalogController < ApplicationController
     if params[:id].start_with?('00000')
       params[:id] = params[:id][5, params[:id].length]
     end
-    params[:id] = "Source " + params[:id]
+    params[:id] = "#{RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE} #{params[:id]}"
   end
 
   def mei
@@ -179,7 +179,7 @@ class CatalogController < ApplicationController
       :"q.alt" => "*:*",
       :rows => 20,
       :defType => 'edismax',
-      :fq => "+type:Source +wf_stage_s:published",
+      :fq => "+type:#{RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE} wf_stage_s:published",
       :hl => 'false',
       :"hl.simple.pre" => '<span class="highlight">',
       :"hl.simple.post" => "</span>",
--- a/config/sample_application.rb
+++ b/config/sample_application.rb
@@ -73,9 +73,21 @@ module RISM
   # search pages (served by Blacklight, that is, /catalog)
   ROOT_REDIRECT = "/catalog"
 
+  # Muscat has two bibliographic record types: Source and Publication.
+  # Source is for unique copies found in libraries, archives or
+  # museums, and strongly tied to musical manuscripts (ex., 100 tag is
+  # named Composer); Publication can describe any bibliographic work,
+  # and the default templates are oriented to describe academic,
+  # research or secondary literature works.  If you will to use Muscat
+  # as originally intended, keep "Source".  If you want to use it as
+  # institutional or research repository, choose "Publication".
+  # Please note that that this second choice is an ongoing work and is
+  # not finished yet.
+  MAIN_BIBLIOGRAPHIC_RECORD_TYPE = "Source"
+
   # Record ids for each records type
   BASE_NEW_IDS = {
-    :publication        => 0,
+    :publication      => 0,
     :holding          => 0,
     :institution      => 0,
     :liturgical_feast => 0,


app/views/catalog/_show_default.html.erb
```diff
@@ -4,12 +4,17 @@
   This is pretty hackish for now - figure out a better way to do this into the CatalogController! -%>
 
 <%
-	@item = Source.find(@document.id.split(" ")[1]) # hackady-hack! Source 1234567 -> 1234567
+	if RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE == "Source"
```
lpugin, line 7: Isn't it possible to convert `RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE ` to model or class?

fjorba: I tried different approaches before publishing this one. This is the most concise approach I found, because this value can be use as such for sending values to Solr in app/controllers/catalog_controller.rb.  Otherwise, I should have to do a couple of conversions to string in `redirect_legacy_values`  and in `configure_blacklight`.  So I stepped back and tried with a string, and now it is much shorter.

xhero:
If understand correctly, you would like to use blacklight (/catalog) to display publications right? I'm not sure this is the best way, as the catalog controller is very tailored to the source model (i.e. indexed marc fields), so probably a better solution would to have a new one and we could configure what /catalog points to.

fjorba:
Of the three test Muscats I have, one is using Sources (IFMuC) and the other two Publications (DDD and Traces), and this trick works, sort of.  Certainly the brief records are too brief for our taste (see #991), and the indices are not adapted to it, but my impression is that it can be worked out.

However, I confess that I feel quite insecure when evaluating whether a new controller is needed or not; I'm still a Rails apprentice, so probably your criterium is more solid than mine.

Indices for publications have to be improved, as now they are specific for musical sources, but I was heading to discuss if Muscat should use the Traject gem (https://github.com/traject/traject) for creating them, as it knows all Marc21 semantics, it has been extensively tested with millions of records of the largest american universities catalogs using Blacklight, and handles all the punctuation and particularities you can think of.  That should be another issue, of course, but I think we'd better goal for adopting Traject and just adapt for each model (Sources, in this case, as the _exotic_ one). I don't know if this project should rethink whether a new model is needed or not.

fjorba:
So, even if you think that are better alternatives, and I'm not discussing this, I still think that this patch is (a) small, (b) solves a specific case (mine, at least), and (c) helps anybody that is interested in evaluating Muscat as research or institutional repository that the things work quite naturally.  And, as a result of (a), it is easily reversible.

xhero:
The biggest problem I see with this approach is that the fields and facets in blacklight are still the source ones, and the variable just switches the filtering of the solr records but leaves the rest (and the set of partials used) the same.

fjorba:
> The biggest problem I see with this approach is that the fields and facets in blacklight are still the source ones, and the variable just switches the filtering of the solr records but leaves the rest (and the set of partials used) the same.

Sure, and that's why I needed to bring Traject (#1110) to light now.

I'm perfectly aware that this is a partial solution.  But it does not incompatible with current Muscat instances (it does not hurt anybody), and it helps new adopters, even if we later need a different controller or solution.

Meanwhile, I can work locally polishing some of the rough edges, but my impression is that with Traject it would be possible to accomodate both bibliographic record types in a single controller (and minimising controllers is one of the things that Jonathan Rochkind says, in its Traject presentation, helps maintanability).

fjorba:
In fact, this is a followup to #958 and #977.

ahankinson:
Given the MARC fields are configurable for each type of thing, can't you change the MARC fields for your "Source" records to keep them as the primary record type, without changing it to the "Publication" type? It seems like you're trying to bypass the built-in Source records because their current configuration doesn't match what you need, but if you simply change the tag configuration for Source to match, then you don't need to do the switching.

The logic of the `if` statement is also a bit off. If I set `RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE = "FlimFlam"`, then it will default to choosing the `Publication` model. (i.e., it doesn't matter what the value of this is, as long as it isn't `"Source"`) This is probably incorrect -- you would want to make the behaviour fit the logic exactly, and make Muscat raise some sort of error if it is doing something unintended. Otherwise if you misspell it, like `Soruce`, then Muscat will happily continue on using the `Publication` model.

fjorba:
I've spend the last year considering different options to use Muscat as a institutional or research digital repository, and I've come to the conclusion that using Publications is much easier and less intrusive than adapting Sources. I've pushed several patches to Muscat with this explicit goal and so far the have mostly favorable accepted.  Moreover, it probably will have some advantages for Muscat itself, like that Publications will be more flexible, and learn things, like having different templates.  I think that the Sources path is not viable.

On the other hand, setting an invalid or misspelled value for RISM::MAIN_BIBLIOGRAPHIC_RECORD_TYPE should not be an argument, because this value is set once per installation, and at this moment less than 5 or 6 persons in the world have this responsability, and of their own good, they won't do such a thing. I haven't put the logic backwards (first Publication, else Source) because I do respect Muscat primary purpose, that is musical sources.

ahankinson:
The critique wasn't about the possibility of misspelling, specifically -- it was about ensuring the logic is well-executed to ensure the result is always as expected. Specifically, this is ['stringly typed' logic](https://cocoacasts.com/the-danger-of-string-literals-and-stringly-typed-code). 

I don't know much about Rails, specifically, but would the `constantize` method be useful here? 

https://apidock.com/rails/String/constantize

fjorba:
Maybe; but probably it is too much of a hassle for a variable that is set once by one person, the admin, an used in probably 5-6 places by 3 or 4 programmers. I confess that I have a preference for minimising requirements and solutions.

xhero:
I think there are several issues here which we should discuss separately.
For BL, my problem here is still that changing that parameter will not necessarily make the controller work or be scalable to use it with another model. For example, when different facets are going to be needed, should we do in the code:

```ruby
if some_selector
  facets_for_selector
elsif some_selector2
  facets_for_selector2
else
  facets_for_default_selector
end
```

I don't think this is the best way to do it - particularly when it is possible to have a dedicated controller. We are also thinking of EOL-ing Blacklight, I've opened a discussion here: https://github.com/rism-digital/muscat/discussions/1113, so we could think of a more clean and scalable solution.

For changing that particular parameter in the catalog_controller, that is something I would do at installation time in this case (and we do it for some other projects to add parameters). But as I said I'm not completely sure the catalog_controller will even properly render Publication models.

I'm also not a huge fan of stringly typed logic too - even if admittedly we use it in Muscat, mostly in legacy code and some other parts, but I'm trying slowly to refactor those.

The last thing is using Publication instead of Sources as the center of the Muscat universe. So Muscat is all built around the Source model - which represents a single musical source, and links to all of its associated data. This is for design and all the application is more or less structured in this way - in particular the database relationships. Now there is linking between the other authority files too, but it is all made keeping Source as the center of the dataset.
It is surely possible to fill in Publications and leave Sources empty but I'm not sure if it is a good solution to morph Publications in a different-named and duplicate Source model, and then have to bolt into muscat a mechanism to switch between the two. The patches submitted were greatly appreciated and actually useful to fix many issues we had, but it was not clear to me that Publication would be an alternative to the Source model. While it is probably doable to be properly implemented it would require quite some thought and engineering.

ahankinson:
It also should probably be understood that "Source" is a legacy from when Muscat was a manuscript-only database. "Central bibliographic record" is probably how "Source" should now be understood -- everything hangs off of Sources, and sources can be manuscripts or prints. There's nothing music-specific about a source, except in how it's configured to meet specific RISM needs.

If you're using "Publications" simply because the name aligns better with the thing that you want to catalogue, then you'll probably end up twisting Muscat in all sorts of ways. As far as I can tell (and @xhero can correct me here!) they're really only meant as bibliographic references attached to things -- they are how you might construct a citation list of publications that talk about a specific source or person. They're not supposed to be a generic model in parallel to Sources.

xhero:
Yup that is exactly the problem! But naming apart I also see @fjorba's side of the problem, to completely adapt Sources for other needs it requires modifying some hardcoded things we have there, we could open a discussion to brainstorm on this?

xhero:
Here we go: https://github.com/rism-digital/muscat/discussions/1115

BTW I'm not changing the name of the model from Sources :wink:

ahankinson:
Awww... `CentralBibliographicItem` just trips off the tongue so easily, though. :-D

fjorba:
I'm impressed about the tsunami I've caused.  And it was only a new variable...

Seriously now, yes, I think that what you explain makes sense. The more and more I have been advancing with the Publications model, the more and more I have been duplicating code with Sources.  Of course, it was not in my hands even to think that you could unify them in a single one, but is a logical step.

If our experience with Invenio serves, yes, it is possible to have both types of bibliographic records in a single software application. The only real difference between sources (unique items, archive-like) bibliographic model and publication is the existence or not of holdings (852) tag, because 856 can coexist with or without 852. And, maybe, the name of the author vs composer (even professor!) field (but not the tag, that would be 100).  Some tags make sense in for one record type and the other in the other one, but they are not incompatible.

So, yes, let's tranfer the discussion to #1115 and do whatever you think with this issue.  I still need it for my systems, and for hypotetical newcomers, and you may or may not accept as interim solution.

ahankinson:
That’s the thing: Sources are also used by us to catalogue prints. So it’s not even used by us for only unique archive items. There is a whole ‘holding’ setup to help deal with this.

fjorba:
So, as the issue has taken another flight, let's close this one.

PR #1111, opened 2021-05-20
Configurable Single Sign-on

tramuntanal:
This PR will add SSO to Muscat.
Related to #1116 

By default, Muscat will still use Devise's `database_authenticatable` but now will allow to configure  other authentication strategies. But it happens that `database_authenticatable` and `cas_authenticatable` are incompatible, so both can not be enabled together.


Gemfile
```diff
@@ -103,6 +104,7 @@ gem 'activerecord-import', ">= 0.4.0"
 
 gem 'ruby-prof'
 group :development, :test do
+  gem 'byebug'
```
ahankinson, line 107: This seems to be introducing a new dependency on `byebug` in an unrelated PR. I would suggest removing it for this contribution.

app/controllers/application_controller.rb
```diff
@@ -25,7 +25,7 @@ def prepare_exception_notifier
   end
 
   def auth_user
-    redirect_to "/admin/login" unless (user_signed_in? || RISM::ANONYMOUS_NAVIGATION || request.path == "/admin/login")
+    redirect_to "/admin/login" unless (user_signed_in? || RISM::ANONYMOUS_NAVIGATION || request.path == "/admin/login" || request.path == "/admin/service")
```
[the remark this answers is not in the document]
tramuntanal: What do you mean by "configurable value"?

A setting like `RISM::REQUIRE_LOGIN_UNLESS= -> { user_signed_in? || RISM::ANONYMOUS_NAVIGATION || request.path == "/admin/login" || request.path == "/admin/service" }` ?

Gemfile
tramuntanal: This dependency will only be applied on :development and :test environments. Bundler will not install the gem if the environment is production (which is automatically set by capistrano).
As it is a usual development tool we added to improve development peace of the project. The weight of the gem is 82.5 KB which doesn't seem a problem for development envs.

Don't you think it's worth it? is it the case that this application has special constraints that we don't know?

app/controllers/application_controller.rb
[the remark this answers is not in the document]
xhero: What does '/admin/service' link to?

Gemfile
ahankinson: I see a few problems:

1. We have no idea of how byebug works, or what difference it makes to running the tests.
2. This dependency is added “on the side” of changing something completely different.
3. There is no documentation on why it was added, other than a very small commit message.

I have no idea of whether it’s a worthwhile addition or not, but I think it should probably be taken out of this PR. If you think it is a worthwhile addition, please propose it in an issue so that we can know more about the benefits.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
tramuntanal: It was needed by the cas_authenticatable gem, but we're removing it, as is incompatible with database_authenticatable, in favor of saml_authenticatable. We will probable have to add some saml related path here. @xhero

xhero:
Should we wait until `saml_authenticatable` is used?

fjorba:
Yes, and probably my tests.

fjorba:
I have tested this patch using UAB CAS SSO server v 6.1 (https://apereo.github.io/cas) and it works very well!

tramuntanal:
Finally, the PR is ready for review.

lib/generators/muscat/templates/devise_saml_authenticatable.rb
```diff
@@ -0,0 +1,77 @@
...
+    # assertion_consumer_service_url is required starting with ruby-saml 1.4.3: https://github.com/onelogin/ruby-saml#updating-from-142-to-143
+    settings.assertion_consumer_service_url     = "https://local-muscat-server.example.org/admin/saml/auth"
+    settings.assertion_consumer_service_binding = "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
+    settings.name_identifier_format             = "urn:oasis:names:tc:SAML:2.0:nameid-format:transient"
+    # settings.authn_context                      = ""
+    # settings.idp_slo_service_url                = "https://remote-sso-server.example.org/simplesaml/www/saml2/idp/SingleLogoutService.php"
+    settings.idp_sso_service_url                = "https://remote-sso-server.example.org/test/saml/idp/auth"
+    # settings.idp_cert_fingerprint               = "9E:65:2E:03:06:8D:80:F2:86:C7:6C:77:A1:D9:14:97:0A:4D:F4:4D"
+    # settings.idp_cert_fingerprint_algorithm     = "http://www.w3.org/2000/09/xmldsig#sha1"
+    # settings.sp_entity_id                       = "https://remote-sso-server.example.org/admin/saml/metadata"
```
fjorba, line 74: I'm testing it and I've just noticed that this line should be this way, and without comment (in our tests with CAS server 6.x needed it):

```
settings.sp_entity_id                       = "https://local-muscat-server.example.org/admin/saml/metadata
```

Sorry for not noticing it before, @tramuntanal.

tramuntanal: @fjorba I've just applied your comment

lib/generators/muscat/templates/devise_saml_authenticatable.rb
```diff
@@ -0,0 +1,77 @@
...
+  # be appended to the named route.
+  # If saml_route_helper_prefix = 'saml' then the new_user_session route becomes new_saml_user_session
+  config.saml_route_helper_prefix = 'saml'
+
+  # You can add allowance for clock drift between the sp and idp.
+  # This is a time in seconds.
+  # config.allowed_clock_drift_in_seconds = 0
+
+  # idp_metadata_parser = OneLogin::RubySaml::IdpMetadataParser.new
+  # config.saml_config= idp_metadata_parser.parse_remote "https://remote-sso-server.example.org/saml/idp/metadata"
```
fjorba, line 62: If I understand well the role of this file, as a template that generates the config/initializers/devise_saml_authenticata\
ble.rb, I think that this file should be without comment, and this way:

config.saml_config = idp_metadata_parser.parse_remote "https://remote-sso-server.example.org/cas/idp/metadata"

I would change /saml/ to /cas/ because we have tested it with CAS server, and at least it will provide a good template for us.

ahankinson: CAS is the name of a specific product, but SAML is the actual standard.

xhero:
Done! Many thanks, we can include it in 8.0 next month, I tested it and it seems to be all ok.

PR #1112, opened 2021-05-21
Fix crontab instructions

fjorba:
Current crontab installation instructions are not following the standard
crontab command procedures.  Adapt them to conventional use.

### DIFF
--- a/2- INSTALL.rdoc
+++ b/2- INSTALL.rdoc
@@ -260,20 +260,26 @@ Also make sure that if there is a proxy, it is configured in `config/initializer
 
 === Crontab installation
 
-In version 7.0 the crono scheduler was removed and a normal crontab plus a script is provided. The default crontab file should be linked in /etc/conn.d. Edit the muscat_crontab to set PATH_TO to the correct muscat installation, you will need to do it by hand.
+In version 7.0 the crono scheduler was removed and a normal crontab plus a script is provided.
 
-On debian/ubuntu this should work:
+To edit your crontab table, become the user where you execute Muscat (via sudo or su www-data, or normal user if you prefer the apache-itk option), and from the shell, execute:
 
- ln -s config/muscat_crontab /etc/cron.d
- chmod +x /etc/cron/muscat_crontab
- sudo service cron restart
+ crontab -e
+
+That will open your $EDITOR and then paste the contents of the sample crontab file provided as config/muscat_crontab.  After saving the file, crontab will check your syntax and give you hints if it finds errors.
 
 If your crontab still has this line:
 
  0 2 * * * cd $PATH_TO && $PATH_TO/bin/rake blacklight:delete_old_searches[7] RAILS_ENV=production
 
-It is save to delete it as it is now taken care but the global muscat crontab.
-*NOTE* If you ever change the cron file name, it CANNOT contains dots! Or it will silently fail.
+It is safe to delete it, as it is now taken care but the global muscat crontab.  To review your table, execute:
+
+ crontab -l
+
+Please take extra care not to execute crontab without any parameter, because then it expects input from the terminal and removes previous contents.
+
+If you execute multiple instances of Muscat in the same server, being real or virtual, via apache-itk or virtual machine, please take care of not executing them at the same hour or day of the week.
+
 
 == Some MySQL optimizations
 
--- a/config/muscat_crontab
+++ b/config/muscat_crontab
@@ -1,12 +1,3 @@
-# Crontab for Muscat
-# Please copy in /etc/cron.d/
-# NOTE since cron is brain dead, this file name MUST NOT contain a dot '.'
-# or cron will not execute it and say nothing. It is documented crazyness
-# as usual so thanks for this crazyness
-SHELL=/bin/sh
-PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
-
-# NOTE
 # PLEASE SET THE CORRECT PATH_TO
 
 # Example of job definition:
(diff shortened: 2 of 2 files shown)


fjorba:
crontab in Linux follows the same Digital Unix or Solaris procedures that I've used in previous decades, so I'm pretty sure that the ones I provide in this patch are the standard ones.

xhero:
I'm confused why you want to change this? The crontab in Muscat is made so it is not installed using the user crontab (which you edit with crontab -e) but it is explicitly installed as a file in /etc/cron.d, which it documented behaivour and also quite widely used. Also having it this way permits to select the user for each command, which in this case is www-data but it does not necessarily need to be, and we can keep the commands in one single place instead of buried around different users.
As a side note, for changes like this I would prefer that we discuss them beforehand, because not all the choices made in Muscat are made up on a whim and some actually do have a reason why they were chosen in the first place.

fjorba:
As I see that I didn't discuss it beforehand, and I jumped in, I accept my mistake and please accept my apologies. I'm sorry, it is my fault. (1/2)

fjorba:
I had reading your instructions as a side when trying to follow your patches, and they seemed strange to me, as I never had to do those low level movements in any of the Unix or Linux systems I've been working with for so many years.  And today, as I'm installing a new test installation following current develop branch, I've reviewed again, I tried to follow, and cannot, as I have a Linux server with several users, and each of them has its own cron table.  The recommended approach I've submitted is a way of saying: system: take this table, and please acommodate it with all other tables you have for all other users.  This way, I understand that they are more general purpose than the original ones, that can only work if the server only has one user that performs cron jobs.

I don't remember instructions like the ones that you provided, and as it contains sentences like how cumbersome this procedure is, I understand that they were due to lack of previous experience. And as I do have it (not like Rails, where I'm a novice), I thought I could give a friendly hand. (2/2)

xhero:
Sorry I did not want to come off a too stern in my reply - but I was a bit surprised to see a patch like that.
So the reasoning behind it is mostly keeping the installation of Muscat neat and ordely, which is easy since Muscat basically exists in its own directory. But for cron jobs to run you need to modify the system wide configuration, and you basically have two choices:

- /etc/crontab
- crontab -r (user crontab)

The global one works well but it can be easily a pain to maintain, since there can be stuff added by the system or other installations. The users are convenient, but they are "hidden" away in /var/spool/cron/crontabs/, each for each user.
When installing a software like Muscat I want that the changes I make to the system are the most clear and reversible, and also the most protected from other installations or manual work.
On Debian/Ubuntu cron also has these directories:

`/etc/cron.hourly, /etc/cron.daily, /etc/cron.monthly, /etc/cron.d`

which are very handy to install package-wide crontabs. So to have a Muscat specific crontab you can copy it over in /etc/cron.d, restart cron, and it will execute. Then you don't have to worry for it to be overwritten or to have different conflicting crontabs (`sudo -u www-data crontab -e` and the `sudo -u other-user crontab -e` and you forget which is which). The caveats are two:

- You need to change PATH to point to the current Muscat installation (but if you have an installation script it can do that for you)
- The name cannot contain a "."

Ans then, all the Muscat-related stuff is just self contained in one file which is quite easy to find and modify or delete.

fjorba:
I understand. Trying to follow the reasoning behind your explanations and how they are different from mine, I think that they differ due to having different expectations or conventions.

You follow the system-wide package system that most Debian packages have, where they add and remove files there when installed or uninstalled. 

Myself, I see Muscat like yet another user package, minimising as much as possible changing users with sudo or similar. In the test server I have, there are 3 different Invenios and 3 different Muscats, plus a couple of local, custom built services, and sometimes new test services are added or removed.  With apache-itk, each one is the owner of their own files, and of course of its own cron jobs.

As both approaches have their logic, if you prefer,  I can redo my patch and, instead of removing your text, and add my proposal as al alternative to it, like the apache-itk one.

fjorba:
And yes, my path to the current version is via the ~ for the user and symbolic links for the current version, so I can have the same for all users (except for the timing, that is):

<pre>
00  08  *  *  sun ~/muscat/muscat-current/bin/muscat_execute_job ModificationDigestJob development :weekly
00  20  *  *  *   ~/muscat/muscat-current/bin/muscat_execute_job ModificationDigestJob development :daily
</pre>

ahankinson:
If you use a deployment system like [Ansible](https://www.ansible.com) then maintaining separate crontab files for different installations is pretty easy, and you don't need to configure it in the "main" code repo (muscat, in this case). 

I would be in favour of keeping the current instructions. If you have local requirements that dictate you do otherwise, then it's OK to have some local documentation setting out the specifics of your own installation.

fjorba:
I don't think adding Ansible as requirement or recommendation to run Muscat would simplify things, as was my purpose. For people new to Rails and Muscat, as I am, we do have enough to learn in order to run it.

My goal was to understand Muscat instructions and maybe help newcomers that may find them as surprising as I did. I recognise that I didn't understand upstream logic, and I did make a mistake deleting the original text. But probably adding them as a complement or alternative may help.

ahankinson:
There is no requirement to use Ansible, either suggested or implied. It was simply a recommendation that you can manage application configuration independent of the application itself to deal with the varying local considerations, and there are tools to help you with this. And you can have local documentation to deal with how you install and manage the application.

Insisting that everyone who uses the application also has to use the same methods for configuring it will lead to a lot of churn in the repo as we try to figure out whose method is "best."

fjorba:
> Insisting that everyone who uses the application also has to use the same methods for configuring it will lead to a lot of churn in the repo as we try to figure out whose method is "best."

If my answer was understood this way, it was my fault, sorry. _Au contraire_, I'm suggesting now to have both instructions so they can help different type of admin types; but mainly newcomers. Expert people need less help, and they can read beyond what is written and adapt the instructions to their uses.

> And you can have local documentation to deal with how you install and manage the application, which will document how your procedures vary from the 'standard' installation.

Of course; but I'm mostly thinking on less experienced people. There is always people with less (and more) experience, and they are the ones that need more detailed instructions.

xhero:
Well we could add a small section in the docs detailing how you can install it with crontab -e, even if I'm a bit wary to what happens when you have a non-login user, so users can choose the method they prefer, and I can remove the header from the crontab ;)

ahankinson:
In my experience the thing that most confuses newcomers is giving two ways of doing the same thing, particularly if they do not have the know-how to decide which one to choose to best meet their needs.

fjorba:
But adding Ansible add

> Well we could add a small section in the docs detailing how you can install it with crontab -e, even if I'm a bit wary to what happens when you have a non-login user, so users can choose the method they prefer, and I can remove the header from the crontab ;)

I confess I have no real-world experience running production systems using a non-login user. I could not run any of my systems with a shell and all utilities the Unix/Linux ecosystem provides; among them, `crontab -e`. And that's why I always use apache-itk.

fjorba:
> In my experience the thing that most confuses newcomers is giving two ways of doing the same thing, particularly if they do not have the know-how to decide which one to choose to best meet their needs.

Certainly, unless you say something like: in case of A, do this; if B, better do that.

fjorba:
My patch proposed an alternative that can be done without my hints, and as upstream is happy with the current instructions, let's close it.

PR #1117, opened 2021-06-03
Update labels for co-composer

ahankinson:
See rism-digital/translations#21 for details. This updates the local translation files for Muscat to bring them inline with the upstream changes, as well as updating the key for the places where it is used.


xhero:
Well spotted! It is weird that it works in Muscat, I think Muscat never crashed on the hyphen. Shall I merge it right away?

ahankinson:
Yep! You can also merge the translations repo PR -- I just did the change on muscat directly to save you needing to update the locale files.

xhero:
super thanks+

PR #1160, opened 2021-11-16
Allow more options for import_from_marc.rb

### LINKED ISSUE #1157: import_from_marc.rb should allow more command-line options
Currently, import_from_marc.rb only allows three positional arguments: filename, record type (or model) and an optional record offset.

The script would be more useful in two aspects: first, allow classical command-line flags and parameters, and also to allow different record importing scenarios.  This would be my draft proposal, that keeps backward compatibility with current syntax and default values:

```
$ rails runner housekeeping/import/import_from_marc.rb --help
Import marcxml records into Muscat database

Usage:  bin/rails runner housekeeping/import/import_from_marc.rb [options]

Options:
 -f, --file        input file name, required
 -t, --type        record type; choose one of: Source, Publication, Holding, Person, Institution, Work
 -m, --from        first record to import
 -v, --versioning  update records version
 -u, --authorities create (scaffold) related Marc authorities records
 -i, --insert      only add new records, skip duplicates
 -r, --replace     only overwrite existing records, skip new ones
 -a, --append      only append non-existing tags to existing records
 -n, --dry-run     only simulate, do not update database
 -h, --help        this help

This script can also be run with positional arguments:

 bin/rails runner housekeeping/import/import_from_marc.rb filename type [from]
```

At this moment I already have the parameter parsing mechanism up and running, but I haven't yet implemented the behaviour of the _new_ values.  Most of them are inspired in the ones we have in classical Invenio, and we depend on them for our regular importing tasks, as we import a great deal of our records.

I'd appreciate your comments and ideas before opeining the corresponding PR. Thanks

fjorba:
Add a standard command-line flags arguments, a help message, check their
validity and add new options.  Fixes #1157.


fjorba:
This patch is less invasive that it seems, as many of the differences are mostly whitespace due to some different conditions that affect indenting.

I have updated my older first patch with another push and I had to resolve some conflicts that I didn't expect. It is my first time with this situation, so I hope that I have resolved it well. If not, please bear with me, and I'll do my best to survive git incantations.

housekeeping/import/import_from_marc.rb
```diff
@@ -1,27 +1,126 @@
...
-# third is the offset to start from
-
-valid_models = ["Source", "Publication", "Holding", "Person", "Institution",
-                "Work"]
-valid_models.each do |model|
-  PaperTrail.request.disable_model(model.constantize)
+# Import Marcxml records into Muscat database
+# Run as bin/rails runner housekeeping/import/import_from_marc.rb
+
+VALID_MODELS = ["Source", "Publication", "Person", "Institution"]
```
ahankinson, line 4: Why is Work removed here?

fjorba: Because I haven't tested them. And I'm unsure about how are they related to Sources, so I prefer to remove them from my first version. They can be added again in the future.

fjorba: I have extensively used it to import Sources, Publications, Person and Institution records, so those are the ones in the list.

ahankinson: But the original script had them, and we certainly use these models in RISM.

fjorba: Or I'll do it myself, of course.

fjorba: Done, thanks for the review and noticing it, @ahankinson!

fjorba:
Hi, @xhero; this patch has been patiently waiting for almost two years. I've been using it almost daily. Is there anything else I can do to being merged? Thanks.

xhero:
For some reason I was convinced you did not want this merged anymore! It looks ok to me.

fjorba:
Grazie!

PR #1162, opened 2021-11-19
Rename revue_title to journal in Publications

fjorba:
Part of #1028 and #1104


fjorba:
Please note that in this commit there is no db/schema.rb due to a local
mismatch. Please add in after performing 'rails db:migrate'. Thanks.

xhero:
Thanks!

PR #1167, opened 2021-11-29
Normalise full_name_d removing diacritics and final punctuation

### LINKED ISSUE #1165: Should full_name_d be further cleaned up?
While loading test Marc records into a local Muscat instance, I've come across several cases that have me thought that maybe further cleaning may be useful.  Marc (sub)fields often end with punctuation, for example:

https://catalog.loc.gov/vwebv/staffView?bibId=1623186

```
1001_ |a Kaczyński, Tadeusz.
24510 |a Lutosławski / |c Tadeusz Kaczyński.
260__ |a Warszawa : |b Sutkowski Edition Warsaw, |c [1994?]-
300__ |a v. <1 > : |b ill., ports. ; |c 25 cm.
440_0 |a Historia muzyki polskiej ; |v t. 9
504__ |a Includes bibliographical references (v. 1, p. [243]) and indexes.
5051_ |a Życie i muzyka
60010 |a Lutosławski, Witold, |d 1913-1994. 
```

Loading records like that into Muscat would create names like that as full_name_d:

```
kaczyński, tadeusz.
lutosławski, witold,
```

Which brings an obvious improvement to remove final punctuation.  But also, maybe the cleaned up versions should remove diacritics, as it is costumary, for those using Latin alphabet, to consider them equivalent (furthermore, as not everybody has had an adequate keyboard when the name was typed).  This is easily done with I18n.transliterate (https://www.rubydoc.info/gems/activesupport/2.3.17/I18n.transliterate), but as noted, only works with Latin alphabets.

That would bring https://github.com/rism-digital/muscat/blob/develop/lib/marc_person.rb get_full_name_and_dates to be something like:

```diff
diff --git a/lib/marc_person.rb b/lib/marc_person.rb
index 586844c..4c0db9a 100644
--- a/lib/marc_person.rb
+++ b/lib/marc_person.rb
@@ -10,40 +10,41 @@ class MarcPerson < Marc

     if node = first_occurance("100", "a")
       if node.content
-        full_name = node.content.truncate(128)
-        full_name_d = node.content.downcase.truncate(128)
+        full_name = node.content.truncate(128).sub(/[ ,\.]+$/, '')
+        full_name_d = I18n.transliterate(full_name.downcase)
+        full_name_d = full_name.downcase if full_name_d.include? '???'
       end
     end
```

```ruby
names = ["Bartók, Béla", "Martinů, Bohuslav", "Lutosławski, Witold", 
         "Schönberg, Arnold", "Schoenberg, Arnold", "Шостако́вич, Дми́трий Дми́триевич"]
names.each { |name| puts "#{name} -> #{name.downcase} => #{I18n.transliterate(name.downcase)}" }
```

With non-latin alphabets, like Cyrillic, we revert it to downcase, as I haven't found an easy alternative:

```
Bartók, Béla -> bartók, béla => bartok, bela
Martinů, Bohuslav -> martinů, bohuslav => martinu, bohuslav
Lutosławski, Witold -> lutosławski, witold => lutoslawski, witold
Schönberg, Arnold -> schönberg, arnold => schonberg, arnold
Schoenberg, Arnold -> schoenberg, arnold => schoenberg, arnold
Шостако́вич, Дми́трий Дми́триевич -> шостако́вич, дми́трий дми́триевич => ???????????, ???????? ???????????
```

It is not perfect (like the Schoenberg ~ Schönberg case), but I think that it would be better than the current transformation.  Moreover, if it can be used for sorting, as in this moment names starting with a letter with a diacritic go at the end of the list.

Would you accept a PR like that?

fjorba:
Documentation for Person model states that full_name_d should be normalised
to downcase and with diacritics removed, but currently it isn't so.  This
patch uses Rails I18n.transliterate, with the known limitation that it only
works for Latin alphabets.  As for other alphabets it returns a string full
of question marks, this patch reverses this damage using the current simpler
downcase normalisation.  Fixes #1165.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -151,6 +151,7 @@ class Person < ApplicationRecord
     new_marc = MarcPerson.new(File.read(ConfigFilePath.get_marc_editor_profile_path("#{Rails.root}/config/marc/#{RISM::MARC}/person/default.marc")))
     new_marc.load_source true
     
+    self.full_name.sub!(/[ ,;\.]+$/, "")
     new_100 = MarcNode.new("person", "100", "", "1#")
     new_100.add_at(MarcNode.new("person", "a", self.full_name, nil), 0)
     
--- a/lib/marc_person.rb
+++ b/lib/marc_person.rb
@@ -10,8 +10,9 @@ class MarcPerson < Marc
 
     if node = first_occurance("100", "a")
       if node.content
-        full_name = node.content.truncate(128)
-        full_name_d = node.content.downcase.truncate(128)
+        full_name = node.content.truncate(128).sub(/[ ,;\.]+$/, "")
+        full_name_d = I18n.transliterate(full_name.downcase)
+        full_name_d = full_name.downcase if full_name_d.include? "???"
       end
     end
     


fjorba:
Please note that (obviously) this patch doesn't fix already existing full_name_d fields in the database.

fjorba:
I'm closing it as it is addressed in #978.

PR #1175, opened 2021-12-13
Add a to_h method to Marc tags to access its subfields

fjorba:
Allow Marc subfields to be accessed like a hash, so it is easy to extract
all subfields of a tag or check the existence or not of a subfield.
Optionally, values can be returned as a list (as_list boolean parameter, for
repetable subfields), and final punctuation can also be optionally stripped
(strip_punctuation boolean parameter).

### DIFF
--- a/lib/marc_node.rb
+++ b/lib/marc_node.rb
@@ -489,7 +489,35 @@ class MarcNode
     end
     return out
   end
-  
+
+  # Convert subfields to hash, optionally as list or stripping punctuation
+  def to_h(as_list: false, strip_punctuation: false)
+    subfields = {}
+    content = looked_up_content if looked_up_content
+    if @tag =~ /^\d\d\d$/
+      if @tag >= "010"
+        @children.each do |child|
+          field = child.to_s
+          if field.start_with? "$"
+            subfield = field[1]
+            if strip_punctuation
+              value = field[2..].sub(/[ ,;\.]+$/, "")
+            else
+              value = field[2..]
+            end
+            if as_list
+              subfields[subfield] ||= []
+              subfields[subfield] << value
+            else
+              subfields[subfield] = value
+            end
+          end
+        end
+      end
+    end
+    subfields
+  end
+
   # Set the foreign object
   def foreign_object=(object)
     @foreign_object = object
--- a/lib/marc_node.rb
+++ b/lib/marc_node.rb
@@ -501,7 +501,7 @@ class MarcNode
           if field.start_with? "$"
             subfield = field[1]
             if strip_punctuation
-              value = field[2..].sub(/[ ,;\.]+$/, "")
+              value = field[2..].sub(/[ ,;:\.]+$/, "")
             else
               value = field[2..]
             end
--- a/lib/marc_node.rb
+++ b/lib/marc_node.rb
@@ -509,7 +509,7 @@ class MarcNode
               subfields[subfield] ||= []
               subfields[subfield] << value
             else
-              subfields[subfield] = value
+              subfields[subfield] ||= value
             end
           end
         end


fjorba:
Its use can be exemplified by #1176.  I'd really appreciate to be included in upstream Muscat, because I consider that accessing subfields as a hash simplifies a lot information extraction, checks and general tag and subfields manipulation.

xhero:
I have nothing against extracting the subtags as hash, but I think the to_h should only do that and not mix different functionalities. I don't think punctuation stripping should be included in the Marc classes as this is really application specific and should be taken care from the part of code that needs this functionality. Not that I'm against having accessors like this, but I think this particular case should be handled in the appropriate controller or in the indexer, that should adapt the data to their needs. Or better even if data should not have punctuation it should be stripped at the import stage.
How does the "list" export work? It seems that "normal" to_h only returns one result if there are repeated subfields, whereas "list" will return an array, is this correct? I think the interface should be coherent in this case, either always return an array (with 1 element when there are no empty fields) or return a string if there is only one element, and an array when there are multiple. If not you risk forgetting `as_list = true` and then pruning away multiple values.

fjorba:
The `as_list` and `strip_punctuation` boolean parameters are both optional and default to false, so there is no chance of _forgetting_ them.

When `as_list` is true, it returns _always_ lists, with one or more elements.  The `strip_punctuation` is a convenience parameter (and optional!) to be used when needed, because when comparing values, sometimes is important to compare them with final punctuation stripped, because the punctiation may depend on the existence or not of a $d or a $e subfield (with authors it happens often).  And as it is so often the case (at least I can attest it), having it at hand adding an `strip_punctuation: true` is really handy.

xhero:
But if I understand correctly how your code works, setting `as_list` to `false` will return only 1 subfield if there are multiple, no? So I think there should be only one way to access the data in a manner that it always returns all the subfields in a consistent way, or else this code will silently remove all repeatable subfields.

While I agree that it is handy to strip the punctuation in this function, it is really not the good place to do it, because it is dependent on the business logic of your application. Putting is here works, but it renders unclear the separation between the marc library (marc_node.rb) and the actual code that uses it. A better way to implement this is in the actual controller that needs it, in a function that explicitly does this.

All this said I think the function name should make it clear that it is returning the values of the subtags, and not the subtags themselves, something like `subtag_values_to_h`, because the accessor functions return marc_node objects and to_h would hint that you are getting a hash of marc_nodes.

fjorba:
> But if I understand correctly how your code works, setting `as_list` to `false` will return only 1 subfield if there are multiple, no? So I think there should be only one way to access the data in a manner that it always returns all the subfields in a consistent way, or else this code will silently remove all repeatable subfields.

Having all values have the same class helps when applying the same methods, like `start_with?`, for example.  If sometimes it returns strings and other returns arrays, it is difficult to apply the same methods.  Many times one is sure (like in #1176), that values are not repeteable, and it helps that they are always strings.  When unsure, the `as_list` parameter ensures that the result consitingly are arrays
 
> While I agree that it is handy to strip the punctuation in this function, it is really not the good place to do it, because it is dependent on the business logic of your application. Putting is here works, but it renders unclear the separation between the marc library (marc_node.rb) and the actual code that uses it. A better way to implement this is in the actual controller that needs it, in a function that explicitly does this.

I'm already using in a batch script for importing records, I'm not sure it the MCV applies there.

> All this said I think the function name should make it clear that it is returning the values of the subtags, and not the subtags themselves, something like `subtag_values_to_h`, because the accessor functions return marc_node objects and to_h would hint that you are getting a hash of marc_nodes.

Here I'm a little bit lost, because I don't quite understand Muscat terminology for subtag.  Marc talks about tags, indicators and subfields.  What is a subtag?

fjorba:
I'm really interested to have such a method in Muscat, and I have reasons to defend the current form:

Regarding the name, it replicates an already existing Ruby method  https://www.rubydoc.info/stdlib/core/Array:to_h, member of the others `to_*` methods. And although it is true that it currently it only turns subfields into a hash, I'd like to make the indicators this way as well.  However, as i feel `marc_node.rb` logic difficult to follow, I still haven't found the way.  In fact, this very difficulty of understanding `marc_node.rb` methods make my proposal more appealing; first, for myself, that I feel a familiar interface to extract data from Marc records, but belonging to standard Muscat as well; but also maybe for other newcomers that may feel like myself.

Regarding the optional parameters, they are optional, and at the same are so tightly bound with the manipulations normally done that makes sense, to me, to have as optional parameters.

We import approximately 2/3 of our records from different sources, most of them in an unattended way, either via cron or ad-hoc.  So, I need ways to check all data from incoming records, whether it exists or not in our database.  It can be the authority records (that's the purpose of #1176) or bibliographic ones (I need to check for strong identifiers, like ISBN, Orcid or Scopus id, or soft ones, like matching bibliographic data).  My current prototype relies on local external indices, because I understand that they are still prototypes and are in flux. But what I need is having hooks in standard Muscat so I can reliably build my tools.

However, if my reasons haven't convinced you, I'll change the name or remove the those optional parameters.

fjorba:
For the time being, I'm dropping this request, as I'm exploring other solutions without interfering Muscat proper.

PR #1176, opened 2021-12-13
Add get_other_standard_identifiers to Marc person

fjorba:
Extract other standard identifiers from Marc person 024 tag, and present
them as a hash, indexed by $2, so it is easy to extract its VIAF, ORCID, or
any other standard identifier using its identifier as key.


fjorba:
For the time being, I'm dropping this request, as I'm exploring other solutions without interfering Muscat proper.

PR #1177, opened 2021-12-13
Remove comments attribute from Person model and db

### LINKED ISSUE #1168: Remove obsolete comments attribute from Person model
Person attributes are created from Marc authority fields.  Currently, Muscat takes person comments from 680 $i:

https://github.com/rism-digital/muscat/blob/c801afabd99b14b25b10609bd29ecbae15fc20d5/lib/marc_person.rb#L71

This field is as reasonable as $a, see:

https://www.loc.gov/marc/authority/ad680.html

However, it turns out that editing a Person authority record, the 680 field creates a single $a subfield.  So, as only 680 $a is used, probably the person model attribute should take it also from $a.

fjorba:
Fixes #1168

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -10,8 +10,7 @@
 # * <tt>composer</tt> - 1 =  it is a composer
 # * <tt>source</tt> - Source from where the bio info comes from
 # * <tt>alternate_names</tt> - Alternate spelling of the name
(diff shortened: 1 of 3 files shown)


fjorba:
Please note that I haven't include db/schema.rb, because as I develop using SQLite, I don't have Innodb and other MySQL/MariaDB specific attributes.

xhero:
Thanks!

Issue #1196, opened 2022-01-24
It is not possible to distinguish or select institutions with $b


fjorba:
If an institution has subordinate units ($b) there is no way to distinguish them in the standard list (maintenance or select dialog).  

![example](https://user-images.githubusercontent.com/1407473/144230814-f7cc1dd4-dade-472f-ae93-a805d531ca4b.png)

I wrote a proposal for a solution in #1171. May I know if it is accepted or viable? If I start working on a patch, will it be merged?

Thanks

PR #1198, opened 2022-01-25
Exclude linked $0 fields from import everywhere

HirschSt:


fjorba:
I have been working during the last weeks on an (external) script to identify the right authority record for persons (#1176) or other authorities (#1175), using external indices, and assigning the right $0 to them.  I our case it is needed as we import many records (around 2/3), many of them with Orcids or university department ids (#1196), but the form is not the same we have in our database.  I was rewriting the input records with the local database authority form (given an authorized id) but also adding a $0 to make sure there is no ambiguity in the assignment.

Which is the current algorithm to assign an authority record the right id?  I'm afraid that without the input $0 would be no certainty about the assignment.

A possible workaround would be to create a command-line option (#1160)

ahankinson:
I believe this change only affects the VIAF import. The MARC records that are being stripped are the responses from the VIAF interface, which is triggered whenever a user tries to import a VIAF record from the admin interface.

xhero:
Yes this is only for VIAF import!

fjorba:
Oh great. I misunderstood the _everywhere_ word, sorry.

xhero:
Wrong button sorry

PR #1284, opened 2022-09-27
Sync admin code for Person and Institution

fjorba:
Muscat offers few more funcionality for Person than Institution, like RISM::CLEVER_ORDERING.  This patch does not add or remove any functionality; it only reorders a few lines and normalizes whitespace so those improvements can be added more cleanly.


fjorba:
Should I open an issue to explain this patch, @xhero? I haven't done it because there is no bug, but I can describe the need, if you prefer.

PR #1287, opened 2022-10-17
Implement $a and $b in institutions

### LINKED ISSUE #1196: It is not possible to distinguish or select institutions with $b

fjorba:

This patch addresses both the authority records attributes and indices in Muscat with the following principles in mind:

* Keep the changes as minimal as possible.
* Do not change Muscat current behaviour if no $b is used in any institutional field.

The patch keeps the current name attribute, but now it is the concatenation of $a and $b (not just $a as before), and adds a new one just for $a (called corporate_name), and another for just $b (called subordinate_unit), that need a migration.

This patch adjustes yaml config files for marc institution, person and source.

Closes #1196 following the discussion at #1171

### DIFF
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -196,6 +196,7 @@ class Institution < ApplicationRecord
 
     # std_title
     self.name, self.place = marc.get_name_and_place
+    self.corporate_name, self.subordinate_unit = marc.get_corporate_name_and_subordinate_unit
     self.address, self.url = marc.get_address_and_url
     self.siglum = marc.get_siglum
     self.marc_source = self.marc.to_marc
--- a/config/marc/tag_config_institution.yml
+++ b/config/marc/tag_config_institution.yml
@@ -309,7 +309,11 @@
     - - a
       - :occurrences: "?"
         :foreign_class: ^0
-        :foreign_field: name
+        :foreign_field: corporate_name
+    - - b
+      - :occurrences: "?"
+        :foreign_class: ^0
+        :foreign_field: subordinate_unit
   "856": 
     :master: u
     :indicator: "1#"
--- a/config/marc/tag_config_person.yml
+++ b/config/marc/tag_config_person.yml
@@ -144,9 +144,13 @@
         :foreign_field: id
         :no_show: true
     - - a
-      - :occurrences: "1"
+      - :occurrences: "?"
         :foreign_class: ^0
-        :foreign_field: name
+        :foreign_field: corporate_name
+    - - b
+      - :occurrences: "?"
+        :foreign_class: ^0
+        :foreign_field: subordinate_unit
   "550":
     :master: a
     :indicator: "##"
--- a/config/marc/tag_config_source.yml
+++ b/config/marc/tag_config_source.yml
@@ -657,9 +657,11 @@
     - - a
       - :occurrences: "?"
         :foreign_class: ^0
-        :foreign_field: name
+        :foreign_field: corporate_name
     - - b
-      - :occurrences: "*"
+      - :occurrences: "?"
+        :foreign_class: ^0
+        :foreign_field: subordinate_unit
     - - c
       - :occurrences: "*"
     - - g
--- a/db/migrate/20221007081534_add_corporate_name_and_subordiate_unit_to_institutions.rb
+++ b/db/migrate/20221007081534_add_corporate_name_and_subordiate_unit_to_institutions.rb
@@ -1,0 +1,6 @@
+class AddCorporateNameAndSubordiateUnitToInstitutions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :institutions, :corporate_name, :string
+    add_column :institutions, :subordinate_unit, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -164,6 +164,8 @@ ActiveRecord::Schema.define(version: 2021_05_14_043225) do
     t.text "alternates"
     t.text "notes"
     t.integer "lock_version", default: 0, null: false
+    t.string "corporate_name"
+    t.string "subordinate_unit"
     t.index ["siglum"], name: "index_institutions_on_siglum"
     t.index ["wf_stage"], name: "index_institutions_on_wf_stage"
   end
--- a/lib/marc_institution.rb
+++ b/lib/marc_institution.rb
@@ -10,7 +10,11 @@ class MarcInstitution < Marc
 
     if node = first_occurance("110", "a")
       if node.content
-        name = node.content.truncate(128)
+        name = node.content
+        if node = first_occurance("110", "b")
+          name += " #{node.content}" if node.content
+        end
+        name = name.truncate(255)
       end
     end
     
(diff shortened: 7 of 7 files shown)


fjorba:
Like my comment to @jenniferward in #1288, this patch does not affect Muscat installations that do not use $b in institutions. It just prepares the software and database to make a sensible use of it.

fjorba:
Hi, may I help to clarify any doubts, or correct any issues about this PR?

xhero:
@jenniferward can you have a look at this? The problem I see here is that we would still need non-repeatable name, while you want to have a repeatable corporate_name and subordinate_unit where linking right?

fjorba:
In our case, we are happy to make $b non-repeatable, to keep things simple.

fjorba:
@xhero, as this patch needs a migration, could you include it for 9.0, please?

jenniferward:
Sorry I am a bit lost - does this, and #1196, just affect the display like when you are looking at https://muscat.rism.info/admin/institutions?

fjorba:
This patch solves #1196, yes, but to solve it and have the $b subfield on is own, I have to create a new field in the database. As each database change deserves a major version (or so I understand), I was hoping to have it before 9.0

jenniferward:
Thanks - Don't we have the $b on its own already? Or do you mean something else?
image

fjorba:
> Why not putting $b together with $a in the name db field? That way we do not need to add a new db field just for $b. (It is a standard practice in Muscat to have several marc subfield gathered together in one db field.)

Because when in an institution where there are hundreds of subordinate units, like an university (see #1196), it is important to $a and $b them separate. This is, moreover the standard cataloging practice (https://www.loc.gov/marc/authority/ad110.html).

My proposal does not affect RISM catalog, not RISM usage or Muscat, but allows us to use it in a way compatible with our catalog, our cataloging practices and the standard usage.

lpugin:
> it is important to $a and $b them separate

I understand, and I know they are hundreds of fields / subfields in marc that we do not use in Muscat, all being related to standard cataloguing practice.

The point is that you can still have $a and $b separate, but then simply display them in one column
```
University A - Unit 1
University A - Unit 1
University A - Unit 1
University A - Unit 1
```

> My proposal does not affect RISM catalog, not RISM usage or Muscat

I think it does because we will have in Muscat an empty column in the institution list, won't we?

fjorba:
> Thanks - Don't we have the $b on its own already? Or do you mean something else?

I explained some of the current limitations of Muscat dealing with institutions at #1171 and #1196. I don't remember the full details of the current behaviour, because I did it a couple of months ago and all my 3 test Muscats are patched this way, but I think that the $b didn't propagate to the authority record or vici-versa.  Moreover, we are interested that Person authority records have 510, so professor A belongs to $a University B $b Department C.  My proposal deals with $b in all Marc records.

fjorba:
The overview of my proposal, including screenshots of the behaviour, is documented in https://github.com/rism-digital/muscat/discussions/1171#discussioncomment-3893998

jenniferward (approved):

xhero:
I finally had a moment to thoroughly review this, but I see a couple problems.
For example in the source tag configuration, you change 710 $a to point to `corporate_name`, instead of `name`. In our data, this will empty all the 710 tags since `corporate_name` is not used and the data is not migrated there. Shouldn't this field still point to `name`, which will contain $a + $b (if there is $b)? In this case it will change nothing for our data and you will still see `corporate_name` + `subordinate_unit`. Besides, for us it is a major change since we would need to re-save all the (millions of) records pointing to institutions.

fjorba:
Thanks for your review, @xhero.  Yes, I saw this problem and wrote my proposal in https://github.com/rism-digital/muscat/pull/1287#issuecomment-1321837162.

If I understand it well, after applying the patch and so, updating the yaml files, you sould update only the institutions table, that, in your case, according to what I see in https://opac.rism.info/main-menu-/kachelmenu/data, is less than 28.000 records. 

The "simpler" sql sentence could just be: `update institutions set corporate_name = name;`. That's usually very fast, a few seconds for this size of records.  I need more time to design a migration that, I recognize, is the _proper_ solucion.  Of course, if you think that I need to write one, I'll do it.

However, I'm convinced that $a must have its own attribute, like $b, because they have to be referenced as such ($a or $b) in the `foreign_class` and `foreign_field` in the tag_config_*.yml fields, as my patch shows.

xhero:
One moment, I'm a bit confused, if we do `update institutions set corporate_name = name;` we are just copying over the data from name to `corportate_name`. In that case, can't you just keep it as it is, with $a that goes in `name`?
If I understand correctly your patch, you need three fields on the DB: `$a`, `$b` and `$a + $b`. Is it right? In this case I think the most simple solution is:

$a -> name
$b -> subordinate_unit (newly added)
$a + $b -> another column (newly added which I think we will not use at RISM, something like `collated_names`?)

In this case, you don't really need to change anything in the code, just add a link to $b where you need it. I think we can live with the DB column name being `name` vs `corporate_name` since it is only an internally used identifier.

This is unless you want to show `$a + $b' in the _autocomplete select boxes_ in the linked models, i.e. `Source`, which would still show the contents of $a (i.e. name). But from your patch it would be the same no?

fjorba:
I'm afraid that your does not cover the display of the full name ($a + $b) in the institutions list, as illustrated in https://github.com/rism-digital/muscat/discussions/1171#discussioncomment-3893998

And yes, we need autocomplete to cover both the $a and the $b, as show in the same link above.  We import 2/3 of our records, not all of them come normalized enough, and so librarians must find all possible forms of the name.  In that sense autocomplete is wonderful to normalize entries.

My proposal was the simplest I could think of that covers both needs: to display the full name in the institutions list and to be able to have $a and $b separated. The only price is to duplicate one field.

xhero:
Ok I see now. I think then the steps should be something like:

* Rename `name` to `corporate_name` and make the foreign classes point to that
* Add a new, properly-named field for the aggregated name
* Make the institution list show that instead of `name`/`corporate_name`

What do you think? This way all the fields are properly named, and it will work as in your patch.

fjorba:
Sorry, but I'm unable to see a real difference between my proposal and yours (and that would be great) if your "properly-named field for the aggregated name" is `name`, and so we avoid a rename (https://github.com/rism-digital/muscat/pull/1104#issuecomment-1365752432).   But I may be missing something.

xhero:
It is for a couple reasons actually. Since `corporate_name` will contain the data we have in `name` at the moment it makes more sense to rename the column instead of creating a new one and then moving the data there. This also creates a "discontinuity" in the project history: there will be one migration that shows us what happened instead of having to guess why in `name` there is different data now.
Besides this I'm not completely sold with the idea of keeping $a and $b just in `name` and the field in this case IMHO it should be more descriptive like `aggredated_name` or something else. It also makes it clear that it is not `name` anymore.

If I just push the patch as it is, since it changes Source 710 to point to `corporate_name` and `corporate_name`, being a new field, will be empty, it will invalidate all the data that we have in 710 Source right now, about 300k records, not counting all the other authority files that are changed.

fjorba:
I see. Please allow me a few days to prepare and test it.

fjorba:
There is a wrong commit here (force-pushed) that I'm trying to remove, sorry.

fjorba:
I've implemented an alternative version, using `full_name` instead of `name` and renaming the attribute and table, in #1326.

However, it is larger and has a couple of limitations, that I explain in #1326. IMHO, this v1 implementation is better, in the sense that is simpler and works as expected, unlike v2.

However, if you @xhero think that it is worthwhile to pursue with v2, I'll try harder to make autocomplete work there. Thanks.

fjorba:
Superseded by #1335.

PR #1300, opened 2022-12-06
Adjust labels and configuration

lpugin:
* Adjust configuration for works
* Add labels for works and work nodes


config/locales/marc_records/en.yml
```diff
@@ -394,6 +394,7 @@ en:
     dedicatee: Dedicatee
     department: Department
     depositor: Depositor
+    deprecated_fields: Obsolete fields (do not use)
```
ahankinson, line 397: These are from the translations files https://github.com/rism-digital/translations/blob/main/locales/en.yml

config/locales/marc_records/en.yml
```diff
@@ -763,6 +764,7 @@ en:
     scoring_in_movement: Scoring in movement
     scoring_summary: Scoring summary
     scribe: Scribe
+    secondary_catalog_works: Secondary catalog of works
```
ahankinson, line 767: Here too.

config/locales/marc_records/en.yml
xhero: Yes I will update the translations repo by hand when we have all the new ones

ahankinson: ok

PR #1315, opened 2023-01-09
Improve DIGITAL_OBJECT_PATH documentation with an example

### LINKED ISSUE #1314: Setting DIGITAL_OBJECT_PATH is underdocumented
The hints in `config/sample_application.rb `about how to configure `DIGITAL_OBJECT_PATH` are scarce, and it is not easy to come by with a combination of paths and symlinks that work.

fjorba:
Configuring DIGITAL_OBJECT_PATH can be tricky due to the behaviour of Ruby on Rails relative and absolute paths.  This patch gives a working example, with hints, on how to create a working one, out of Muscat software directory tree, so it is not affected by changes on it.

Closes #1314

### DIFF
--- a/config/sample_application.rb
+++ b/config/sample_application.rb
@@ -95,9 +95,21 @@ module RISM
   #   43200 = half a day)
   VERSION_TIMEOUT = 43200
 
-  # Set the path for the digital object storage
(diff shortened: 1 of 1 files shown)


xhero:
thanks!

PR #1319, opened 2023-01-11
Recognize subunit as different from monograph in sources leader

### LINKED ISSUE #1318: When importing printed music sources, all of them goes to edition_content
No matter if the bibliographic level in the music leader is d (Subunit) or m (Monograph/Item), Muscat puts all them as subunit (edition_content).

https://www.loc.gov/marc/bibliographic/bdleader.html

fjorba:
Marc21 bibliographic level distinguishes between d (subunit) and m (monograph/item) that, in Muscat record types, correspond to edition_content (Printed music edition, item in collection) and edition (Printed music edition, collection parent record or individual item).

https://www.loc.gov/marc/bibliographic/bdleader.html

Split the logic between both types instead of putting all together as edition_content.

Closes #1318

### DIFF
--- a/lib/marc_source.rb
+++ b/lib/marc_source.rb
@@ -244,8 +244,10 @@ class MarcSource < Marc
       rt = RECORD_TYPES[:collection]
     elsif leader.match(/......d[dm].............../)
       rt = RECORD_TYPES[:source]
-    elsif leader.match(/......c[dm].............../)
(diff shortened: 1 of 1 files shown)


xhero:
Thanks for the patch, I would like to merge it after @HirschSt has a chance to review it since he is the one that normally manages imports and I want to make sure it is ok for him.

PR #1326, opened 2023-01-30
Implement $a and $b in institutions, v2

fjorba:

Following @xhero request in #1287, this second version renames the old `name` attribute to `full_name`, while keeping `corporate_name` for $a and `subordinate_unit` for $b.  This attribute rename forces changes in other files, so it is larger and more invasive than the first one

This version has two limitations or drawbacks:

While testing it, I kept finding and "undefined method `name' for many hours of unsuccessful debugging, the only cure was an "alias_attribute :name, :full_name", which, I know, only hides the problem.

The second one is that autocomplete for institutions fields does no longer work, and I haven't been able to find why.

### DIFF
--- a/app/admin/institution.rb
+++ b/app/admin/institution.rb
@@ -22,10 +22,10 @@ ActiveAdmin.register Institution do
   #
   # temporarily allow all parameters
   controller do
-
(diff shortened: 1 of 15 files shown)


fjorba:
Superseded by #1335.

PR #1335, opened 2023-03-13
Implement $a and $b in institutions, v3

fjorba:

Following @xhero request in #1287, this third version renames the old name attribute to full_name, while keeping corporate_name for $a and subordinate_unit for $b.  This attribute rename forces changes in other files, so it is larger and more invasive than the first one.

Regarding v2, this version:

* no longer needs an alias_attribute :name, :full_name
* autocomplete now works for institution fields
* no longer modifies Muscat default editing templates

I have been testing it and I hope this time is correct.


xhero:
Thanks! I think it is ok, I will do a review asap and merge!

PR #1357, opened 2023-04-14
Develop csv

HirschSt:


db/schema.rb
```diff
@@ -120,6 +130,20 @@
     t.index ["wf_stage"], name: "index_holdings_on_wf_stage"
   end
 
+  create_table "holdings_institutions", id: false, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb3", force: :cascade do |t|
```
ahankinson, line 133: Maybe I'm missing something, but we already have a `holdings_to_institutions` table...

db/schema.rb
```diff
@@ -170,6 +194,20 @@
     t.index ["wf_stage"], name: "index_institutions_on_wf_stage"
   end
 
+  create_table "institutions_institutions", id: false, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb3", force: :cascade do |t|
```
ahankinson, line 197: Same here... there is already an `institutions_to_institutions` table...

db/schema.rb
```diff
@@ -170,6 +194,20 @@
...
+    t.integer "institution_a_id"
+    t.integer "institution_b_id"
+    t.index ["institution_a_id"], name: "index_institutions_institutions_on_institution_a_id"
+    t.index ["institution_b_id"], name: "index_institutions_institutions_on_institution_b_id"
+  end
+
+  create_table "institutions_to_catalogues", id: false, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb3", force: :cascade do |t|
```
ahankinson, line 204: What's a catalogue? There is no `catalogue` table, so I'm not sure what table `catalogue_id` is supposed to point to...

db/schema.rb
```diff
@@ -37,6 +37,16 @@
     t.index ["user_id"], name: "index_bookmarks_on_user_id"
   end
 
+  create_table "crono_jobs", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb3", force: :cascade do |t|
```
ahankinson, line 40: Looking at it, I'm not sure these migrations are necessary for the CSV export.

ahankinson (changes requested):

HirschSt:
It would be even more interesting why my schema is different ...

ahankinson:
I believe `catalogues` were renamed to `publications`, and quite a few join tables were renamed from `institution_institution` to `institution_to_institution`. This happened a year or so ago. 

Perhaps you are running a much older database structure?

HirschSt:
I will check this later and prepare an new pr, thanks :monocle_face:

PR #1358, opened 2023-04-17
Develop csv

HirschSt:


lpugin:
Why all these commits where there is only one file change?

HirschSt:
I'm afraid this is the result when you have the schema.rb file in version control of git. Only real solution would be to have it in .gitignore, which is not recommended. Any proposals?

ahankinson:
You can rebase your changes in git on the latest develop.

 - Copy your changes to the CSV writer
 - Delete this branch in your local repo
 - Switch to the develop branch in your local git and pull in all changes, including all the changes to schema.rb. This should put your `develop` in sync with GitHub
 - Create a new `develop_csv` branch off the develop branch
 - Re-add your changes back to the CSV writer
 - Submit a new PR

There is probably an automated way to do that with some fancy git command, but that's the most straightforward way.

HirschSt:
Thanks, but this is my workflow. Nevertheless I have to run the migration (with changes to schema.rb).

ahankinson:
The issue is that you've commited, then undone, all the changes, which means this branch will commit a lot of unnecessary changes to the main develop branch that ultimately bring it back to the same point. Your changes to `schema.rb` won't actually run as far as I can tell, since you have wiped them out.

If you re-start with a 'clean' history you will then only have a single commit with your changes. Your additions, and then deletions, to `schema.rb` will not appear anywhere in the history of the project.

PR #1363, opened 2023-04-25
Accept both 240 and 245 for publications title, prefer the later

fjorba:
Secondary literature should change title tag from 240 to 245 (#626).  As a transition step and to keep compatibility with current records, accept both, but prefer 245.

Accept subtitle as $b if exists, as well; it does not affect sites that do not use subtitles.

Partially addresses #626


xhero:
I need a comment from @jenniferward and @ahankinson on this

jenniferward:
I can't really picture what this change looks like so my comments probably won't help. In RISM Muscat we don't yet have 245 in Publications. We also don't have $b (nor do we need it for now, I'd say), but I agree that any project that wants to use 245$a or 245$b should be able to.

ahankinson:
@jenniferward you did file the original issue for this, way back in 2018. Does this still fix it, or have things moved on since then?

https://github.com/rism-digital/muscat/issues/626

jenniferward:
Nothing has changed since 2018. But what are we looking at? Does this mean there will be a new field, 245?

xhero:
Yes, the idea is to move all 240s to 245

jenniferward:
Then that's great from my point of view.

HirschSt:
Please have in mind that this will break existing exports

fjorba:
> Please have in mind that this will break existing exports

Not yet. I mean, not until you change your RISM database records. But we, at UAB, are customizing secondary literature template, and want to use 245 for title. My patch ensures that the title element for the publication model works both for (your current) template with 240 and (our testing) template with 245. And when in the future you change your template, this step will already be done.

fjorba:
Hello, @jenniferward, if I understand it the current status of this PR, misses your approval. As I wrote, my proposal doesn't hurt any Muscat instance. If a publication has 240 as title (as current RISM practice), it takes it as such, but if it has 245, it will accept it. If a record has both, it takes 245. And if is has subtitle in $b, it recongizes is as such; if not, it doesn't hurt. Thanks.

jenniferward:
Right yes, sorry! I approve. Anyone who uses Muscat should reasonably expect 245$a and $b to be present and function as described by @fjorba .

jenniferward (approved):

fjorba:
Hi @xhero, may I ask if this patch is going to be accepted, or should i drop it, and why? Thanks.

xhero:
Sorry this flew under my radar, but I think we are all in _favour_ actually! We can make official just after I survive Muscat 11.

fjorba:
Hi @xhero, as you have fortunately survived Muscat 11, maybe now you could merge this patch, that has been waiting a year and a half without objections. Thanks.

PR #1378, opened 2023-04-27
Add support for duplicating a work

lpugin:
Based on similar implementations

### DIFF
--- a/lib/marc_work.rb
+++ b/lib/marc_work.rb
@@ -69,4 +69,8 @@ class MarcWork < Marc
     return status
   end
 
+  def reset_to_new
+    first_occurance("001").content = "__TEMP__"
+  end
+
 end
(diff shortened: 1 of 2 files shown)


lib/marc_work.rb
```diff
@@ -69,4 +69,8 @@ def get_link_status
     return status
   end
 
+  def reset_to_new
+    first_occurance("001").content = "__TEMP__"
```
ahankinson, line 73:
```suggestion
    first_occurence("001").content = "__TEMP__"
```

ahankinson: Although since this is a function call it probably needs to be changed elsewhere as well?

fjorba: It happens often, yes:
$ git grep first_occurance | wc -l
164

PR #1390, opened 2023-05-24
Fix current folder filter behaviour and create a couple more

### LINKED ISSUE #1389: There are too few filters to select from the list of folders
Currently there is a single folder filter to select from the folders list, by name. Moreover, the title ("Any name contains") does not correspond to the real behavior, that is, the full name.

When there are many folders, this selection is too limiting. The current "contains" should really do a substring select, not a full name, and other filters, like by owner and folder type, would be very welcome.

fjorba:
Make sure that the current filter for "contains" does a substring search, not a full name search, and create a filter by owner and another for folder type.

Closes #1389

Thanks to @ivan-mr from @CodiTramuntana for his help with procs and collections.

### DIFF
--- a/app/admin/folder.rb
+++ b/app/admin/folder.rb
@@ -176,12 +176,12 @@ ActiveAdmin.register Folder do
   filter :wf_owner_equals, :label => proc { I18n.t(:filter_owner) },
          as: :select,
          collection: proc {
-           if current_user.has_any_role?(:editor, :admin)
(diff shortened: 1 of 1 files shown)


xhero:
Ciao Ferran! I did not forget about you and the PRs, I'm going to try to review them in the next few days.

xhero:
Thanks!

PR #1395, opened 2023-06-01
Develop abilities

HirschSt:


lpugin:
Could you add a description please? Thanks!

HirschSt:
Sure:
- Make id in html-table view of SRU clickable
- Added 240m (scoring) to SRU index
- Allow editors (not only admins) to view users

PR #1397, opened 2023-06-02
Add authority reference counters to secondary literature records

### LINKED ISSUE #1392: Should the number of occurrences in Secondary Literature appear in authority record list?
Currently, in the list view for People, Institution, or any other authority records, there is a column with the number of sources where this entry occurs. However, it says nothing about how often it occurs in secondary literature records.

I suspect that this information would be as useful for the main Muscat instance for RISM as it is for us at UAB.  However, before providing a patch, I'd like to know if you are interested in it.

A different issue is that there is not much space left with the current design of those authorities list as created with the ActiveAdmin gem.  I think that it can optimized, but that will be another issue that I'll open soon.

So, to make it clear: I can provide a patch for creating those counters (it is basically a copy&paste of the sources one, so no black magic here), and it requires a Solr reindex. The fact that those counters appear now or later on the list would be a different patch, maybe related with the optimization of the current list.

fjorba:
Muscat has counters with the relation of sources to authority records, as Solr integer fields.  This patch creates counters also that relate publications (secondary literature) to people, institutions, standard terms, works and work nodes.  Counter for places already existed.  It cannot be created (easily, at least) for standard titles, as currently the database relation for those two tables does not exist.

A full Solr reindex is required after creating those indices (bin/rails sunspot:reindex).

Closes #1392

Visualizing those counters will be added later, probably after some adjustments to the authority record list views.


fjorba:
Please note that I cannot test if it works, or works well, or even if it makes sense, for works or work nodes, as I still don't fully understand them.  So please feel free to correct any absurdity or misbehavior here. Thanks.

xhero:
I finally had a chance to look at this, the counters seem ok, we can fine tune them later. Merci!

PR #1398, opened 2023-06-02
Rename obsolete catalogues filter text to publications

fjorba:
There are still obsolete references to "catalogues" before being renamed to "publications", this time in the filter strings.

Part of #1028


xhero:
Danke!

PR #1419, opened 2023-06-27
Develop sru index

HirschSt:
Because indexing of 100$a in sources was reconfigured, SRU configuration is updated


ahankinson:
There are a bunch of schema changes in this PR as well. Should they be there?

HirschSt:
We can discuss this again on Wednesday meeting

xhero:
Can't you just do
`git checkout -- db/schema.rb`
perfore pushing to the PR?

xhero:
@HirschSt could you please modify the PR so it has no schema.db? Since it does not touch the database I think it does not need to be here in the first place. This way I can merge it and have it in 9.3.1.

lpugin:
@HirschSt it seems that your `schema.rb` is not up-to-date. It still contains some tables that do not exist any more. See `crono_jobs` deleted in [this migration](https://github.com/rism-digital/muscat/blob/develop/db/migrate/20230124101722_delete_unused_crono_jobs.rb)

PR #1420, opened 2023-06-28
Develop sru index

HirschSt:
2nd try :-)


ahankinson:
Hm. Now this will delete the `schema.rb` file entirely. And the commit history goes with it. 

You should pull an up-to-date copy of the develop branch, then make a new, clean branch from it and make your changes to the SRU config.

HirschSt:
ok

PR #1388, opened 2023-05-23
Allow items to be added and removed from folders

fjorba:
Currently the only way to add items to folders is via a batch selection form a list created in the list view.  This patch allows to add and remove them from the full record view.  Closes #931.

Thanks to @ivan-mr from @CodiTramuntana for his help finding a way of doing it following the Rails conventions.

PR #1438, opened 2023-09-04
Allow items to be added and removed from folders, v2

### LINKED ISSUE #931: Add items to folders from full record view
I would like the ability to add items to folders from the full record view.

fjorba:
Currently the only way to add items to folders is via a batch selection form a list created in the list view.  This patch allows to add and remove them from the full record view.  Closes #931.

This second version unifies in a single (and hopefully complete) commit and replaces #1388.

Thanks to @ivan-mr from @CodiTramuntana for his help finding a way of doing it following the Rails conventions.

### DIFF
--- a/app/admin/digital_object.rb
+++ b/app/admin/digital_object.rb
@@ -238,6 +238,7 @@ ActiveAdmin.register DigitalObject do
     if digital_object.images?
       render :partial => "activeadmin/section_sidebar_do_links", :locals => { :item => digital_object }
     end
+    render :partial => "activeadmin/section_sidebar_folder_actions", :locals => { :item => digital_object }
   end
 
   ##########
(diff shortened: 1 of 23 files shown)


xhero:
Very cool thanks! I will have a look soon and we can include it in the next release

xhero:
Could you have a look at the .where in the view? Once it is fixed I can merge this. Thanks!

fjorba:
There are two .where, but I'm unable to see the problem. Could you please which of them and which problem does it has? Thanks

```
<% compatible_folders = Folder.where(folder_type: item.class) %>
<% folders_with_this_item = compatible_folders.where(id: item.folder_items.pluck(:folder_id)) %>
```

xhero:
SQL queries should be done in the controller and not the view, this is to keep the logic separation between the view and the controller.

fjorba:
Done. Thanks again to  @ivan-mr for this solution using a helper.

lpugin (approved):

PR #1442, opened 2023-09-08
Add DEL acronym for deleted records, unify translation files

### LINKED ISSUE #1441: Missing translation code for deleted records, some translations out of sync
Muscat records can have several statuses: published, inprogress, deprecated and deleted.  Depending on the entity, they can be strings, numeric or symbols. There is also an three letter acronym "translation" that is shown in the record list.  Except for the deleted, that does not exist.  At UAB we are interested in having it. We propose `DEL` as the obvious one.

Moreover, according to the English translation file (`en.yml`), and effectively the application only uses those acronyms should exist only for English, and for the first Muscat translations (French, German and Italian), the `status_codes` list does not exist, but the ones that came later do have it, even though it is not used.

fjorba:
Deleted records did not have a translation acronym, and some newer translation files do have translations for status codes, although they are not used.  Remove them.

Fixes #1441

### DIFF
--- a/config/locales/ca.yml
+++ b/config/locales/ca.yml
@@ -868,26 +868,3 @@ ca:
   blacklight_advanced_search:
     form:
       start_over: "Torna a l'inici"
-
-
-######################################################################################################
-# Record types, English only
-
-  record_types_codes:
-    "0": "UNK"
-    "1": "COL"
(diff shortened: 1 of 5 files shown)


xhero:
Thanks!

PR #1468, opened 2024-01-12
Display subfield in Marc editor

fjorba:
When cataloging Marc records, the Marc tag number is shown (ex: 100), but not the subfield (ex: $a).  This information is useful to understand which value has to be filled in, and to find documentation and/or examples.

This patch adds it between parenthesis after its name, like the tag.

### DIFF
--- a/lib/editor_configuration.rb
+++ b/lib/editor_configuration.rb
@@ -136,6 +136,7 @@ class EditorConfiguration
         label = I18n.t(labels_config[id][:fields][sub_id][:label], :locale => :en)  + " [translation missing]"
       end
 
+      label = "#{label} ($#{sub_id})" if edit
       return label
     end
     # if nothing found


jenniferward:
Nice, thanks! This has been a minor annoyance on a subconscious level but I was unable to bring it to the surface (I always clicked on the guidelines to find the subfield).

fjorba:
Let me add a screenshot with the result
subfields

ahankinson:
It would also be possible to format this as a mouse-over, so that the subfield would be displayed as the user hovered the field. That might satisfy both?

fjorba:
Personally, I like to see the things right there, so screenshots work, and finding strings in the page (with Ctrl-F) work as well.  When I have put useful information as a mouse-over, I have found that some people is not aware of its existence.  At UAB, when I added those subfields, our librarians said: Finally! Let's see what @jenniferward and their team say.

jenniferward:
I think our librarians will echo the "finally"! But let's go with it how it is and pay attention to how our users react.

jenniferward:
This has been a huge convenience when editing and discussing records, @fjorba, so thank you for this enhancement. In fact now I (and others) find myself looking for the MARC field when viewing a record. Do you think it would be feasible to add the MARC field to the record's View mode?

fjorba:
I'm afraid I'm not quite familiarized with the Muscat bibliographic view mode. It turns out that we are interested that our catalogers can see exactly what will be shown in our current Invenio systems (and our future Blacklight systems). So I've develop an in-house formatter for our bibliographic records to match them, although we do use standard Muscat views for authority records, MARC or non-MARC.

I'm sure @xhero will do it better and faster thay myself.

What we do also have is a record history view to display the differences comparing MARC, that I briefly mentioned at #1634 (issue that I still have to close appropiately), and will propose in a specific issue or discussion.

PR #1469, opened 2024-01-12
Create links to display Marc records in Marc and Marcxml

### LINKED ISSUE #595: Proposal for full record view
The current record full view / preview options do not display all of the fields. A full display is needed in order for contributors to check their work. Please see the attached document for the arrangement and order of the fields. ZR and CH collaborated on this.

[Proposal preview Muscat CH_ZR an die Entwickler.docx](https://github.com/rism-ch/muscat/files/1695745/Proposal.preview.Muscat.CH_ZR.an.die.Entwickler.docx)

fjorba:
Although Muscat catalog records in Marc, Marc formats are not easily available.  Marcxml is accessible adding .xml at the end of any record, but it is a format to exchange records between systems, not to be read by catalogers.  This patch adds three links to Marc records: one to the default Muscat view, another to a a simple, human readable but complete Marc21 record with all the fields and subfields, and a third on to the Marcxml one.

Maybe it closes #595


fjorba:
The links in the sidebar look like this
marc-views

jenniferward:
Thank you! From a user's perspective, I agree the MARC should be easier to access. Would be interested in testing this against #595 too.

ahankinson:
I find the spaces around the subfield and the value to be problematic, particularly for debugging fields that have errant spaces. Can the regex be adjusted, or simply removed, to not inject those spaces?

fjorba:
This format is to be read by people, and spaces help them to read it better.  If I had to clean errant spaces (as I certainly have done in my life), I would use a program to find and fix them. In our opinion, this corner case should not penalize the daily job of catalogers examining Marc records.  Marcxml is also available to take a second look.

ahankinson:
I don’t know… the record view is the one the humans will mostly look at. Since your patch that adds the subfield to the label, that’s effectively a human MARC view.

The spaces are not actually in the record, so it’s problematic to show them in this view as well.

I would imagine an Important use case for this view is copying and pasting into another piece of software, and there the spaces will definitely cause problems.

ahankinson:
The format we use for these records is the one from MarcEdit. Here's the explanation of how it works.

> MarcEdit’s mnemonic format is a very precise, but simple format.  The key formatting rules to follow are:   Field data must always be on one line.  A new line represents a new field in the mnemonic format.  A blank line is an indicator for a new record. The field format is as follows:  `=100  \\$adata.`  The parts are specific.  All new fields begin with an equal-sign, followed by the field number/tag.  Following the field number/tag, are two blank spaces.  Following the spaces are the indicators.  Blank indicators are presented as ‘\’ or spaces.  Finally, after the indicators, is the field data.  Subfield characters are represented by a $ + character code.

Source: https://marcedit.reeset.net/howto-editing-your-first-recordset

xhero:
I too think we should not add spaces, if then by chance somebody copies and imports the data back, it will have the spaces preserved!
I too think we should use CSS. At some point in history Muscat actually had a MARC view with syntax highlighting!

fjorba:
I have heard about Marcedit, but I have know practical knowlege of it. The format I have followed is the one produced by yaz-marcdump (https://software.indexdata.com/yaz/doc/yaz-marcdump.html), that has been part of Debian since many ages (https://packages.debian.org/yaz) and adds spaces when converting, for example, from marcxml to line marc (using their terminology). Blacklight "librarian view" also add spaces (https://search.library.yale.edu/catalog/646376/librarian_view).

But again, this is less important that that having Marc formats readily available.

I'm not specially good at css and styling.   That said, should I do anything else here?

lpugin:
l am not very convinced by the `"a"` icon for MarcXML. Why not `<xml>`, or `</>` ?

fjorba:
The idea is that how subfields are represented. If the canonical subfield in Marc is show as ` $a` , in Marcxml is ` "a"`, so one mimmicks the other. ` <xml>`  is too wide, but `</>` is a good alternative indeed.

jenniferward:
MARC all in caps?
MARC21
MARCXML

fjorba:
Are you referring to the icons or the text that accompanies them?  Regarding text, I personally prefer lowercase, but certainly the canonical page always uses uppercase (https://www.loc.gov/marc/). Icons should fit the other ones, so MARCXML probably would be too wide, but if they are ambiguous, the text helps to clarify what it does.

jenniferward:
I meant in your screenshot above, the text to the right of the icons. We should probably match the official styling, which is all uppercase.

fjorba:
Yes, you are right.

fjorba:
Second proposal, following your suggestions
marc-colors

I'm using this simple css styling, so that it can be easily changed:

```css
.marc21 {
    font-family: monospace;
}

.marc_tag {
    color: maroon;
}

.marc_subfield {
    color: red;
}

```
If you agree, I update the patch.

ahankinson:
Better! The only comment is that I might suggest making them bold as well, and/or using another colour than red, since red is a common problem for colour-blind people.

fjorba:
Third proposal, using bold both for the tags and subfields, and different colors. I've tried to choose dark colors, because otherwise, it is harder to read. But again, styling is not my forte. I've choosen the colors from the list at https://htmlcolorcodes.com/color-names/

marc-colors-alternative
```css
.marc21 {
    font-family: monospace;
}

.marc_tag {
    font-weight: bold;
    color: saddlebrown;
}

.marc_subfield {
    font-weight: bold;
    color: darkgoldenrod;
}
```

xhero:
I find it _very_ nice!

fjorba:
Thanks for your feedback!

app/views/marc/_show.html.erb
```diff
@@ -9,7 +9,12 @@
 	<%= render :partial => "marc/show_diff_wf_stage" %>
 <% end %>
 
-<% layout_config["group_order"].each do |group_name|
+<% if params[:view] == "MARC21" %>
+    <div class="marc21">
+	<%= @item.marc_source.gsub(/(=[0-9][0-9][0-9])  /, '<br><span class="marc_tag">\1</span> ').gsub(/(\$.)/, '<span class="marc_subfield">\1</span>').html_safe %>
```
ahankinson, line 14:
```suggestion
	<%= @item.marc_source.gsub(/(=\d{3})  /, '<br><span class="marc_tag">\1</span> ').gsub(/(\$.)/, '<span class="marc_subfield">\1</span>').html_safe %>
```

ahankinson:
In the future you can just click "Commit suggestion" and it should incorporate it in the PR.

fjorba:
> Colors look great! I would personally add some CSS spacing since that does make it easier to read.

Sorry, I'm not sure to understand it. Where at? How?

lpugin:
Something like this (not tested)
```css
.marc21 {
    font-family: monospace;
}

.marc_tag {
    font-weight: bold;
    color: saddlebrown;
}

.marc_subfield {
    font-weight: bold;
    color: darkgoldenrod;
    margin-left: 0.8ex;
    margin-right: 0.8ex;
}
```

fjorba:
I see! Yes, now I have my spaces back, but via CSS. Great, thanks!

subfield-margins

PR #1479, opened 2024-01-31
Improvements to Catalan and Spanish translations

fjorba:
Use common constructs.


xhero:
Thanks!

ahankinson:
I think these should be done through CrowdIn? These files are updated from there, no?

xhero:
nope, these are the internal ones!

PR #1512, opened 2024-02-26
Update PersonLabels.yml

jenniferward:
Changing label (key) of Personal names 100$d `Years of birth and death` to `Life dates` to match 100$y and 400$d, as well as 100$d in Sources.


ahankinson:
LGTM but I'll let @xhero merge when he's back.

PR #1519, opened 2024-02-28
Renaming 520 in Secondary Literature

jenniferward:
We had a field in Kallisto to record dates of publication for a journal (to clarify: beginning and end dates of a serial publication). This is in Muscat as 520 but is labeled `Contents note`. This correction renames the field to `Dates of publication` (following the new key added in https://github.com/rism-digital/translations/pull/124).

### DIFF
--- a/config/editor_profiles/default/configurations/PublicationLabels.yml
+++ b/config/editor_profiles/default/configurations/PublicationLabels.yml
@@ -104,8 +104,8 @@
 '520':
   fields:
     a:
-      label: records.contents_note
-  label: records.contents_note
+      label: records.dates_of_publication
+  label: records.dates_of_publication
 '599':
   fields:
     a:


fjorba:
Those Dates of Publication and/or Sequential Designation, shouldn't go to 362 (https://www.loc.gov/marc/bibliographic/bd362.html)?. 520 is clearly for summary, abstract, etc  (https://www.loc.gov/marc/bibliographic/bd520.html).

We have been configuring them in our own yaml configuration files just a few weeks ago, so they are fresh in my memory.

jenniferward:
You are right, but I am in a bit of a bind. See https://github.com/rism-digital/muscat/issues/1490. The problem is that we have data that behaves like MARC 362 but it is in a field called `Contents note` and assigned to MARC 520.  I was hoping that as a first step, I could at least assign the correct label to the field, then the label would at least match the data. 

Over in Sources, we call 520 `Description summary` which is in alignment with MARC.

We don't really need a 520 `Description summary` in Publications in our use of it.

fjorba:
I see, that's a reasonable strategy.

jenniferward:
Sorry, I think I did something wrong!

image

jenniferward:
Ignore my above comment, it's fine now.

PR #1534, opened 2024-03-12
Update Muscat saml SSO authentication to Rails 7

fjorba:
New Rails 7 loader (zeitwerk) conflicts with previous setup, and it does not load the libraries in the order they used to.  Change it slightly, and add an override rule in config/sample_application.rb

All the hard work, and thus, the merit, goes to @ivan-mr from @CodiTramuntana.

### DIFF
--- a/config/sample_application.rb
+++ b/config/sample_application.rb
@@ -187,6 +187,15 @@ module Muscat
     
     # See what was loaded
     #Rails.autoloaders.log!
+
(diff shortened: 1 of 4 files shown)


xhero:
Thanks! I'm a bit behind the review, I will include in the next release after 10.2.

xhero:
Sorry I'm super late here, but looks ok! thanks!

PR #1535, opened 2024-03-15
Add missing user help files for Catalan and Spanish

fjorba:
Without those files, it is not possible to edit user alerts if locale is set to Catalan or Spanish, as the application fails due to a missing help file.


xhero:
Thank you!

PR #1581, opened 2024-05-21
Add display_name to people

fjorba:
The commit 3ff0f26f6f5f161a764802dfc00a2e33c7f7c1ea removed full_name_d, that we have been using to create a richer display form of people (with life dates and other subfields that disambiguate similar names) for lists and autocompletions.  As requested by @xhero, this patch creates a new column named display_name for this purpose.  Its value will be generated in lib/marc_person.rb, with the goal to be generated only once, as it is used quite often, and generate it every time is costly.  I'm leaving this part for a future PR, as I need to make our usage compatible with upstream Muscat.


xhero:
thanks!

PR #1589, opened 2024-05-28
Update wf_owner when non-marc authority records are updated

### LINKED ISSUE #1584: Is wf_owner updated for non-marc records? Should it?
I've just discovered that in my Muscat instances, `wf_owner` is updated only for Marc records (publications, sources, people, institutions, etc), but not for standard_titles, standard_terms, places or liturgical_feasts, even after effectively updating records.  I'm pretty sure that I haven't patched anything related to this, but may I ask if it happens also in your instances, please?

If not, would you agree that it should be updated?

I have seen that all models, Marc and non-Marc, have the same `belongs_to :user, :foreign_key => "wf_owner"` statement, so the reason for not being updated appears to be a bug.

```
MariaDB [traces_development]> select distinct wf_owner from publications;
+----------+
| wf_owner |
+----------+
|        1 |
|        0 |
|        7 |
+----------+
3 rows in set (4.566 sec)

MariaDB [traces_development]> select distinct wf_owner from people;
+----------+
| wf_owner |
+----------+
|        3 |
|        6 |
|        0 |
+----------+
3 rows in set (0.238 sec)

MariaDB [traces_development]> select distinct wf_owner from institutions;
+----------+
| wf_owner |
+----------+
|        1 |
|        0 |
+----------+
2 rows in set (0.073 sec)

MariaDB [traces_development]> select distinct wf_owner from standard_titles;
+----------+
| wf_owner |
+----------+
|        0 |
+----------+
1 row in set (0.015 sec)

MariaDB [traces_development]> select distinct wf_owner from standard_terms;
+----------+
| wf_owner |
+----------+
|        0 |
+----------+
1 row in set (0.008 sec)

MariaDB [traces_development]> select distinct wf_owner from places;
+----------+
| wf_owner |
+----------+
|        0 |
+----------+
1 row in set (0.001 sec)

MariaDB [traces_development]> select distinct wf_owner from liturgical_feasts;
+----------+
| wf_owner |
+----------+
|        0 |
+----------+
1 row in set (0.070 sec)

```

fjorba:
Make this field visible in the display page, and update it using current user field.  After uptading, redirect to the display page, because if redirects to the list, the modified record is lost and not easily accessible, except via the browser history.

Closes #1584

### DIFF
--- a/app/admin/liturgical_feast.rb
+++ b/app/admin/liturgical_feast.rb
@@ -70,7 +70,7 @@ ActiveAdmin.register LiturgicalFeast do
     # redirect update failure for preserving sidebars
     def update
       update! do |success,failure|
-        success.html { redirect_to collection_path }
+        success.html { redirect_to resource_path(params[:id]) }
         failure.html { redirect_back fallback_location: root_path, flash: { :error => "#{I18n.t(:error_saving)}" } }
       end
       # Run the eventual triggers
@@ -140,6 +140,7 @@ ActiveAdmin.register LiturgicalFeast do
       row (I18n.t :filter_name) { |r| r.name }
       row (I18n.t :filter_alternate_terms) { |r| r.alternate_terms }
       row (I18n.t :filter_notes) { |r| r.notes } 
+      row (I18n.t :filter_owner) { |r| User.find_by(id: r.wf_owner).name rescue r.wf_owner }
     end
     active_admin_embedded_source_list( self, liturgical_feast, !is_selection_mode? )
 
@@ -180,6 +181,7 @@ ActiveAdmin.register LiturgicalFeast do
       f.input :alternate_terms, :label => (I18n.t :filter_alternate_terms)
       f.input :notes, :label => (I18n.t :filter_notes)
       f.input :wf_stage, :label => (I18n.t :filter_wf_stage)
+      f.input :wf_owner, :input_html => { :value => current_user.id }, :as => :hidden
       f.input :lock_version, :as => :hidden
     end
   end
--- a/app/admin/place.rb
+++ b/app/admin/place.rb
@@ -69,7 +69,7 @@ ActiveAdmin.register Place do
     # redirect update failure for preserving sidebars
     def update
       update! do |success,failure|
-        success.html { redirect_to collection_path }
+        success.html { redirect_to resource_path(params[:id]) }
         failure.html { redirect_back fallback_location: root_path, flash: { :error => "#{I18n.t(:error_saving)}" } }
       end
       # Run the eventual triggers
@@ -159,6 +159,7 @@ ActiveAdmin.register Place do
       row (I18n.t :filter_country) { |r| r.country }
       row (I18n.t :filter_district) { |r| r.district }    
       row (I18n.t :filter_notes) { |r| r.notes }    
+      row (I18n.t :filter_owner) { |r| User.find_by(id: r.wf_owner).name rescue r.wf_owner }
     end
 
     active_admin_embedded_source_list( self, place, !is_selection_mode? )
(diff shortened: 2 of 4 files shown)


ahankinson:
> After uptading, redirect to the display page, because if redirects to the list, the modified record is lost and not easily accessible, except via the browser history.

@jenniferward should confirm if this is desired behaviour

fjorba:
To clarify my slight change of the behaviour: it all starts with the current Muscat behavior (inherited from ActiveAdmin) that "remembers" the last search for each kind of record, in that case, non-marc authorities. Let's imagine that the last time you were visiting places was Brno. Later on, maybe another day, you see in any record (source, publication, person or institution) a reference to a misspelled Zaggreb. You clik on the authority record for Zaggreb to correct it and suddenly you find yourself a list of cities around Brno. Moreover, if your finger also pressed a key near the return one, you may not be sure that Zagreb had been effectively corrected.  I have bitten by this behaviour.

This slight change makes that after correcting a place, subject, title or liturgical feast, Muscat shows the record that you have just corrected, so you can check if your correction was right or not.

xhero:
Does this change the owner every time a record is saved? For the search problem, in the current devel branch Muscat is tab independent so searches should not cross tabs anymore.

xhero:
> > After uptading, redirect to the display page, because if redirects to the list, the modified record is lost and not easily accessible, except via the browser history.
> 
> @jenniferward should confirm if this is desired behaviour

Maybe we can add two links like in the MARC editor, "save and close" and "Save and stay here"?

fjorba:
> Does this change the owner every time a record is saved?

I'm not sure to fully understand your question, due to the links between bibliographic and authority records.  This patch just corrects the owner of the non-marc authority record (places, subjects, titles and liturgical feasts) you have just updated. Depending on how is well received by you, I have further proposals, some of them already implemented in my installations, but I wanted to start with a simple case.

> For the search problem, in the current devel branch Muscat is tab independent so searches should not cross tabs anymore.

That's good news. However, does it remember "old" searches like the Brno vs Zabreb scenario I've described above? Because I think that seeing what you have already corrected is a good thing on itself.

fjorba:
> > > After uptading, redirect to the display page, because if redirects to the list, the modified record is lost and not easily accessible, except via the browser history.
> > 
> > 
> > @jenniferward should confirm if this is desired behaviour
> 
> Maybe we can add two links like in the MARC editor, "save and close" and "Save and stay here"?

The "Save and stay here" in the MARC records means that you don't leave the editor. In non-Marc records, the current behavior is to leave the record and go to the list of the last searched authority of the current type.

jenniferward:
I agree with the new behavior: after you save a non-MARC authority, you should be taken to the full record view, not the list. "Save and close" should be the same across Muscat - whether an element is MARC or not shouldn't make a difference in terms of this behavior.

Many thanks @fjorba !

jenniferward:
In case this is relevant: I don't see owners for Standard Titles (https://muscat-test.rism.info/admin/standard_titles/50359392).

image

fjorba:
Yes, this is also true: when authority records are created (at least) via importing bibliographic ones (or related authority), those newly created do not have owner.

I was waiting the outcome of this PR to open a new issue or discussion asking about ownership of records, but I prefer not to complicate this simple patch.

xhero:
I finally had a moment to look at this - seems ok to me, auth files can be edited only by editors right?

fjorba:
Any objections here?

xhero:
I still think we should not update the owner of the record to the current user when it is saved

ahankinson:
> I still think we should not update the owner of the record to the current user when it is saved

+1

I don’t want to become a record owner if I simply correct a typo

fjorba:
I see. So, I'll cancel this owner-change feature. Is staying in the full record after updating the record? At least @jenniferward was happy about this behavior. In that case, I'll rewrite the patch.

jenniferward:
Yes, I was focused on "staying in the full record after updating", not the owner-change part (which went over my head).

fjorba:
Partially replaced by #1695.

PR #1601, opened 2024-06-12
Add Saudi Arabia to Institution Country Codes list

ahankinson:
Since we now have a Saudi Arabian institution with a siglum, we should probably also have this as a selectable country code


xhero:
I managed to miss this for like 2 minutes!

Docudoctor:
Please do not forget https://github.com/rism-digital/muscat/issues/1543
Thanks.

Am Mi., 12. Juni 2024 um 16:16 Uhr schrieb Rodolfo Zitellini <
***@***.***>:

PR #1602, opened 2024-06-14
Make logs less verbose on development mode

fjorba:
It is difficult to find useful information in Muscat log files due to the the noise that the workers write.  Some time ago I googled around to find an easy to implement solution to this situation, as it should be common to any Rails app.  I found this one to be simple, and works very well, at least for me.


fjorba:
As I find it useful, may be it is useful also for you. It is part of my stack of local patches that I apply to each release.

PR #1605, opened 2024-06-26
If set to index when importing records, tell Solr to commit them

fjorba:
Without commiting the index, indexing is lost.  After months without understanding why it didn't work, now I see it, yes.


xhero:
merci!

PR #1606, opened 2024-06-26
Add i18n to record actions pulldown texts

fjorba:
Record action text was only in English, so add the i18n wrapper.  Add also the deprecated and deleted statuses, if they are also useful at RISM; otherwise, just drop both lines.


xhero:
Thanks!

PR #1609, opened 2024-07-09
Pregenerate and use a person display_name to disambiguate similar ones

fjorba:
When different people share the same or similar name, other fields are used to distinguish them, like life dates.  In Muscat, this richer form of name is used, at least, in autocomplete forms, although it could be used on other places.  Instead of building it each time, create it once and store it in db.

### DIFF
--- a/app/admin/person.rb
+++ b/app/admin/person.rb
@@ -36,7 +36,7 @@ ActiveAdmin.register Person do
   # temporarily allow all parameters
   controller do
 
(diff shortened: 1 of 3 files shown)


fjorba:
This patch does not address the scenario of empty `display_name` fields when they are not yet created. In my instances, they are already pregenerated during the import phase. Maybe a temporary solution could be to check in `autocomplete_label` if its empty and, in that case, `set_object_fields` and save before doing a retrospective creation with a specific job?

PR #1610, opened 2024-07-09
Remove obsolete utf8=✓ from url parameters

fjorba:
Since Rails 6 (August 2019), it is considered that it is no longer necessary to check for obsolete browsers, so remove it also from Muscat.  This creates shorter and cleaner urls.  See
https://guides.rubyonrails.org/6_0_release_notes.html#action-view-notable-changes


xhero:
Sorry this flew under my radar... thanks!

PR #1612, opened 2024-07-10
Optimize space in users list

fjorba:
Muscat inherits default ActiveAdmin actions (View, Edit, Delete) at the right of its lists.  This patch removes those columns and places its links in other places, as proposed in #1405:

* Delete column is removed, as probably it is safer to delete from the full record rather than from the list.
* View column is linked to the Name column, as it is the "main" one, the one that clearly identifiers the record.
* Edit column is linked to the Id colum, but only if the current user has permission to edit it; otherwise, it just shows its value.  Of course, records can also be edited from the full record.

As a small improvement, date of creation column has been made sortable.

### DIFF
--- a/app/admin/user.rb
+++ b/app/admin/user.rb
@@ -61,14 +61,14 @@ ActiveAdmin.register User do
     selectable_column
     column "Id", sortable: :id do |user|
       if can?(:edit, user)
-        link_to user.id, edit_admin_user_path(user)
+        link_to user.id, edit_admin_user_path(user), { title: I18n.t(:edit) }
       else
         user.id
       end
     end
     column :username
     column :name do |user|
-      link_to user.name, admin_user_path(user)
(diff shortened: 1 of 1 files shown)


fjorba:
I'm aware that nobody has reacted to my proposal at #1405. However, here at UAB we have reorganized all Muscat tables using this principle and we feel that there is more space to work with. Moreover, in very wide screens or tables, the View or Edit table are far away from the record you intuitively would like to point to, as it is customary in web applications.

If this patch is accepted, I will send all the others. The last one will be the person lists, because there is a bolder proposal using `display_name` and removing life dates. But even without that, the View, Edit and Delete links can be supressed in all lists and get more space.

fjorba:
As this proposal had no consensus (see #1405), I'm dropping it and keep as local patch.

PR #1651, opened 2024-10-23
Display record status tag in full record and edit screen

### LINKED ISSUE #1649: Should the record status be more visible in the show and edit screen?
Currently, the record status (`wf_stage`) is quite visible in Muscat in the record lists (index), for example:

image

However, in the full record view, or in the edit screen, the status is only visible in the bottom part of the record, so discreet that, unless you look at the button of the screen, it is not obvious to know if the record you are watching or editing is published or not:

image

We think that the record status is so important that it should be _very_ visible, and if possible, using the same colors and conventions that in the list, that is, reusing the same ActiveAdmin `status_tag`.  For example, in the record view:

image

Or in the edit screen:

image

We have implemented it for all record types that have `wf_status`; and yes, we know that some of them are under discussion (#1600), but it use can be adjusted later.

I'd like to push my patch upstream if you are also interested in it.

fjorba:
Record status is only visible in the record lists screen, but not when the records are shown or edited.  Display it at the top of the screen using the same ActiveAdmin status_tag than in the lists.

Closes #1649

### DIFF
--- a/app/admin/institution.rb
+++ b/app/admin/institution.rb
@@ -51,7 +51,7 @@ ActiveAdmin.register Institution do
       @show_history = true if params[:show_history]
       @editor_profile = EditorConfiguration.get_default_layout @item
       @editor_validation = EditorValidation.get_default_validation(@item)
-      @page_title = "#{I18n.t(:edit)} #{@editor_profile.name} [#{@item.id}]"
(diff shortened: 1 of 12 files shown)


fjorba:
This patch has a annoying behaviour displaying html in the title bar of the browser, at least in Linux, both with Firefox and Chromium and Chrome

image

image

I suspect that it should be fixed in ActiveAdmin, not in Muscat, because the title is sent as such to ActiveAdmin.

xhero:
@jenniferward What do you think about this? Maybe it will help people remember to publish... We can fix the problem with <span> appearing on the title.

PR #1678, opened 2024-11-28
Add initial work node help (en and fr)

lpugin:
* To be reviewed and translated


app/views/admin/work_nodes/_work_nodes_help_show.en.md
jenniferward, line 1: I think more context is needed initially for the Muscat user clicking on "Work nodes" and needing to decide whether it is relevant for their work. I'd start it like this (and am I correct in assuming GND numbers are preferred?). Parts of the text can also be repeated in the guidelines for the specific fields. 
---

The **Work nodes** serve as a link between sources in Muscat and external work authorities. Within Muscat, they group sources sharing the same external work authority ID number. A link to an existing, external work authority, preferably to the GND from the German National Library (DNB), is required to create a work node. 

Work nodes are not exported as authority data, and work nodes do not have external ID numbers. They are linked in the source records through the field **External work reference** (930), and link to the external authority through that authority's work ID number.

lpugin: Please feel free to adjust the text as needed. I'll update the French version accordingly.

xhero:
Is the PR ok to be merged?

jenniferward:
I'd like @BaMikusi 's opinion on the text to ensure clarity.

BaMikusi:
If you ask me, this looks perfect!

PR #1698, opened 2024-12-17
Display wf_stage in place authority records

fjorba:
Of all non-marc authority records, place is the only that doesn't display wf_stage, although the field exists in the database.  Make it visible, so it is homogeneous with the others.

Part of #1584


xhero:
Sorry I missed this! ok for me & thanks!

PR #1728, opened 2025-03-12
Update en.yml

jenniferward:
We call this "Cataloging language" in the guidelines (https://guidelines.rism.info/cataloguing.html#cat_language and throughout) and it refers to what catalogers select in 040$b. The field should be named the same as this phrase..


ahankinson:
This gets changed here: https://github.com/rism-digital/translations/blob/main/locales/en.yml#L588

The files then get copied to Muscat once the translations are made.

PR #1740, opened 2025-03-27
Remove deleted and deprecated status from records

### LINKED ISSUE #1738: Remove "Deleted" and "Deprecated" from Status
Catalogers should not be able to delete records, so "Deleted" needs to be removed from the "Record Status" selector for everyone. (It is misleading because the record is removed from the public catalogs, like "unpublished", but does not disappear from Muscat.)
Image

"Deprecated" should also be removed, at least from our perspective, because we don't know what catalogers think this means, and we can't search for it.

fjorba:
Deleted and deprecated status are not used at RISM. Fixes #1738.

Rewrite current statuses in a more compact form.


fjorba:
Rewritten by @xhero at #1738.

Issue #1251, opened 2022-05-12
Filters applied should be more obvious in the search


jenniferward:
We have this Search Status box under the filters on any Muscat page where you can search:
![image](https://user-images.githubusercontent.com/18663291/168095089-48d81c26-f189-4d93-9d59-7a7d2618094d.png)

but can it be made more obvious that filters are activated? My search of 1745 - 1790 was from yesterday but Muscat retained it when I reopened the page, and I had forgotten about it and was confused for a moment when there were no results. Also, after doing some searches from within Sources to other authorities, the search terms are also retained/carried over and appear in the "Any field" -- again, not at all obvious and can lead to unintentional results. 
My preference would be to move the search status box up to the top and make it blazing red, but maybe there are other ideas.

PR #1838, opened 2025-07-15
Add a clear button at top when a filter is active

fjorba:
When a filter is active, current value is shown at the top of the sidebar (since the #1448 PR).  However, the button to clear the filter is at the bottom.  For some record types with lots of filtering options, this can be quite annoying as the clear button is down under, it forces to scroll down a lot and distracts from the current task.

This patch adds another clear button at top when there is an active filter, to clear it out easily. Complements #1251.

Thanks to @ivan-mr from @CodiTramuntana for the implementation.

Maybe the muscat.scss patch needs some adjustments, as we were working on our own personalized scss for each our sites, and the original styling seems to inherit from upstream activeadmin gem.


jenniferward:
I think users will find this helpful. The placement is logical. I agree that the scrolling can get annoying. Thanks!

xhero:
Sorry I ended up overlooking this! I promise I will try ASAP!

PR #1872, opened 2025-10-06
Normalize extraneous and punctuation characters

### LINKED ISSUE #1599: Normalize punctuation on input
Some characters need to be normalized (smart quotes vs. apostrophes) but some need to be allowed (u vs ü). Currently, `’` and `'` are read as different punctuation marks. This causes misalignment in city names in Institutions:
La Seu d’Urgell https://rism.online/institutions/30079707 
La Seu d'Urgell https://rism.online/institutions/30005481 
and duplicates in Titles/Texts:
https://muscat.rism.info/admin/standard_titles?utf8=%E2%9C%93&q%5Btitle_equals%5D=Au+sein+des+alarmes+l%E2%80%99amour+a+des+charmes&commit=Filter&order=id_desc 
Au sein des alarmes l’amour a des charmes
Au sein des alarmes l'amour a des charmes
 
This arises especially when copying from websites or data imports. The problem has been solved with searching (see  https://github.com/rism-digital/muscat/issues/622 ) but not on the input side.

I can think of the following:
-  `’` and `'` 
- `" "` and `“ ”`
- `-`  `–`  `—` (dash, n-dash, m-dash)

For the dashes, only one is needed (the dash I think?) in the standardized fields.

What about spaces? Sometimes that acts strangely (Excel doesn't always read the spaces as spaces) but I can't describe it further than that.

This is most important the fields that are linked to authority files, not everywhere (like in notes fields).

fjorba:
Change alternative characters to standarized ones, like smart quotes vs apostrophes, but also different kinds of quotes, spaces or dashes.

It can be used also to autocorrect misspellings or characters that are not easy to type in all keyboards.

Closes #1599

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -399,7 +399,7 @@ GEM
     puma (7.0.3)
       nio4r (~> 2.0)
     racc (1.8.1)
-    rack (2.2.17)
+    rack (2.2.18)
     rack-protection (3.2.0)
       base64 (>= 0.1.0)
       rack (~> 2.2, >= 2.2.4)
--- a/app/admin/compare_versions.rb
+++ b/app/admin/compare_versions.rb
@@ -139,6 +139,7 @@ ActiveAdmin.register_page "Compare Versions" do
 
   sidebar I18n.t "compare_versions.options", :class => "sidebar_tabs", :only => [:index] do
     #@arbre_context.assigns[:test] to get variables from content
+    # NOW USE controller.view_assigns["test"]
 
     # no idea why the I18n.locale is not set by set_locale in the ApplicationController
     I18n.locale = session[:locale]
--- a/app/admin/dashboard.rb
+++ b/app/admin/dashboard.rb
@@ -70,7 +70,7 @@ ActiveAdmin.register_page "Dashboard" do
       limit = params[:dashboard_quantity].to_i
     end
 
-    @file = @arbre_context.assigns[:file]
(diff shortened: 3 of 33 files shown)


ahankinson:
Is this for search? Isn’t this done in Solr already?

fjorba:
It may well be. Sure you have a better view of all resources Rails provides. Would you like me to try it as a concern, or do you prefer to adapt it yourself?

xhero:
I'm testing this but I'm not sure it works?
The function returns a string, but the string never seems to be associated to the original db field, or am I missing a step?

fjorba:
Properly, it doesn't need to return anything, because it alters the model.marc_source (for those records with `marc_source`, and it works for me i all my tests. If I type, for example `<<something>>`, I get  `«something»`, and this is what it is stored in the db.

fjorba:
I'm sorry but I don't understand what this 4e14c5f4a75767c3e6bb80a50c880e9e3b558f64 commit is. It is clearly a mistake, I cannot recall how did I do it, and I don't find a way to remove it. Should I remove this PR and push it a new one?

ahankinson:
I have lots of questions about the nature of this change!

There are standardized ways of normalizing Unicode characters -- https://unicode.org/reports/tr15/

This deals with things like how do you normalize characters that can be either one or two code points -- that is, a single character é, or an e with a combining ´ character.

I'm also not sure if we always want to blindly normalize `<<` to «

PR #1979, opened 2026-05-28
Make person identifiers json column compatible both MySQL and MariaDB

fjorba:
Adding the :json attribute to the :identifiers column does not hurt MySQL but allows MariaDB to recognize it as JSON.  See
https://github.com/rails/rails/issues/44997


xhero:
Thank you!

